PR #33, opened 2022-05-16
Issue 7: fix distribution notice

ianturton:
Fixes the distribution licence on the source files https://github.com/lutraconsulting/mergin/issues/7


PeterPetrik:
This is solved now, all files should have correct licencese

PR #35, opened 2022-06-08
fix: removed unnecessary port mappings for Postgres and Redis

GeoSpark:
Redis and Postgres were visible on the public network, which is a big security flaw. This fix removes the port mappings, and references them by their network name rather than their IP address.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,8 +4,6 @@ services:
   db:
     image: postgres:10
     container_name: mergin-db
-    ports:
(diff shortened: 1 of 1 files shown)


PeterPetrik:
Hi @GeoSpark, we published major CE release today, I am sorry we did not manage to review and merge your commit, but from now we will develop Mergin CE directly on GitHub so it will be more smooth from now on. I will close the PR and please redo new one from current master if needed.

PR #57, opened 2023-04-04
Access requests list pagination

harminius:
**Description**

Added pagination for Project access requests endpoints:

- /project/access-requests:
- /project/access-requests/{namespace}

**FE part**

- implement paginated functions for project access requests in Profile and Dashboard
- changes in project's store


server/mergin/sync/private_api.yaml
```diff
@@ -445,6 +477,29 @@ components:
           requested_by:
             type: string
             example: jane.doe
+    ProjectRequestAccessListItem:
```
MarcelGeo, line 480: Just suggestion:

- rename to ProjectAccessRequest (or ProjectAccessRequestItem)
- use it in schema ProjectRequestAccessList (maybe also rename in right order)

```
    # Array of Pets
    type: array
    items:
      $ref: '#/components/schemas/Pet'
 ```

File: server/mergin/sync/private_api_controller.py [hunk 2]
```diff
@@ -127,17 +127,24 @@ def get_project_access_requests():
 
 
 @auth_required
-def list_namespace_project_access_requests(namespace):
-    """List of incoming project access requests to workspace"""
+def list_namespace_project_access_requests(namespace, page, per_page, order_params=None):
+    """Paginated list of incoming project access requests to workspace"""
```
MarcelGeo, line 131: Just suggestion:

Unify with description in private_api.yaml

server/mergin/sync/private_api.yaml
```diff
@@ -26,13 +26,45 @@ paths:
...
+          in: query
+          description: Number of results per page
+          required: true
+          schema:
+            type: integer
+            maximum: 50
+            example: 20
+        - name: order_params
+          in: query
+          description: Sorting fields e.g. expire DESC
```
MarcelGeo, line 47: Can we unify pattern with get paginated projects (v1) ? "expire_desc" ??

varmar05: the one used now is newer (and already used at various places), it would be better to unify it with projects pagination the other way around - but because it is public api it is not that easy - what we can do is start using the new format of order params while keeping old functional and start updating clients

server/mergin/sync/private_api.yaml
```diff
@@ -26,13 +26,45 @@ paths:
...
             application/json:
               schema:
-                $ref: "#/components/schemas/ProjectRequestAccessList"
+                type: object
+                properties:
+                  count:
+                    type: integer
+                    description: Total number of all project requests
+                    example: 20
+                  project-requests:
```
MarcelGeo, line 64: Rename project-requests -> items.

server/mergin/sync/private_api_controller.py
```diff
@@ -127,17 +127,24 @@ def get_project_access_requests():
...
     )
-    return jsonify(ProjectAccessRequestSchema(many=True).dump(access_requests)), 200
+    if order_params:
+        order_by_params = parse_order_params(AccessRequest, order_params)
+        access_requests = access_requests.order_by(*order_by_params)
+
+    result = access_requests.paginate(page, per_page).items
+    total = access_requests.paginate(page, per_page).total
+    data = ProjectAccessRequestSchema(many=True).dump(result)
+    data = {"access_requests": data, "count": total}
```
MarcelGeo, line 146: Rename access_requests -> items

MarcelGeo:
It would be useful to add searching for access requests endpoints based on project_name.

server/mergin/sync/private_api.yaml
```diff
@@ -26,13 +26,21 @@ paths:
           required: true
           schema:
             type: string
+        - $ref: "#/components/parameters/Page"
```
varmar05, line 29: is there any reason pagination params are not merged into single reference?

server/mergin/sync/utils.py
```diff
@@ -404,3 +404,30 @@ def get_order_param(cls: Model, order_param: OrderParam) -> Optional[UnaryExpres
 def is_valid_gpkg(file_meta):
     """Check if diff file is valid"""
     return file_meta["size"] != 0
+
+
+def parse_order_params(cls, order_params):
```
varmar05, line 409: I think this is generic function for any model and it would be better if it is somewhere in top directory not in sync module.

server/mergin/tests/test_private_project_api.py
```diff
@@ -193,6 +207,101 @@ def test_project_access_request(client):
...
 
 
+def test_get_project_access_requests(client):
+    """Test paginated list of project access requests initiated by current user in session"""
+    user = User.query.filter(User.username == "mergin").first()
+    test_workspace = create_workspace()
+    add_user("test_user", "ilovemergin")
+    login(client, "test_user", "ilovemergin")
+
+    # create 10 projects ("user" as an owner) and requests to them from "test_user"
```
varmar05, line 217: 10 or 12?

web-app/packages/lib/src/modules/project/components/ProjectAccessRequestTableTemplate.vue
```diff
@@ -166,30 +178,45 @@ export default Vue.extend({
...
+    async fetchItems() {
+      this.loading = true
+      try {
+        const payload: GetProjectAccessRequestsPayload = {
+          namespace: this.namespace,
+          params: {
+            page: this.options.page,
+            per_page: this.options.itemsPerPage,
+            order_params:
+              this.options.sortBy[0] &&
```
varmar05, line 207: should not we support multi-column sorting?

web-app/packages/lib/src/modules/project/store.ts
```diff
@@ -482,15 +488,15 @@ const ProjectStore: Module<ProjectState, RootState> = {
             `/login?redirect=/projects/${namespace}/${projectName}`
           )
         } else if (e.response.status === 403) {
-          ProjectApi.fetchAccessRequests()
+          ProjectApi.fetchAccessRequests({
+            // TODO: Add searching based on project_name to this endpoint
```
varmar05, line 492: is not is already the case?

varmar05:
@harminius could you please update description so it is clear what was done it this PR?

web-app/packages/lib/src/modules/project/components/ProjectAccessRequestTableTemplate.vue
MarcelGeo: This interface is from vuetify data-table options.

server/mergin/sync/private_api.yaml
MarcelGeo: parameter are not possible to merge to one type: object ref, because there have to be name parameter in PaginatedParams:

```
parameters:
    PaginatedParams:
      in: query
      required: true
      **name: pagination**
      schema:
        type: object
        properties:
          page:
            type: integer
            minimum: 1
            example: 1
          per_page:
            type: integer
            maximum: 100
            example: 50
```

Requests will be instead ?page=1&per_page=1 -> ?pagination.page=1&pagination.per_page=1

PR #62, opened 2023-04-20
Add custom error into definition of create/clone project endpoints

varmar05:
Defines new error type in response of create and clone project endpoints.


harminius (approved):
LGTM

PR #66, opened 2023-04-21
add black and eslint check

PeterPetrik:


MarcelGeo (approved):
We have to fix linting errors after migration to Vue3.

PR #67, opened 2023-04-21
Fix projects access update endpoints

varmar05:
This PR deal with project sharing / access updates:
- reworks public endpoint to support partial update introducing new error framework
- fixes private endpoint for sharing

### DIFF
--- a/server/mergin/sync/private_api_controller.py
+++ b/server/mergin/sync/private_api_controller.py
@@ -285,5 +285,6 @@ def update_project_access(id: str):
         project.access.unset_role(user_id)
     else:
         project.access.set_role(user_id, ProjectRole(request.json["role"]))
+        project_access_granted.send(project, user_id=user_id)
     db.session.commit()
     return NoContent, 200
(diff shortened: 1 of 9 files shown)


server/mergin/sync/private_api_controller.py
```diff
@@ -276,14 +276,15 @@ def update_project_access(id: str):
...
-    if user_id == project.creator_id:
+    if user.id == project.creator_id:
         abort(400, "Ownership of project creator cannot be removed")
 
     if request.json["role"] == "none":
-        project.access.unset_role(user_id)
+        project.access.unset_role(user.id)
     else:
-        project.access.set_role(user_id, ProjectRole(request.json["role"]))
+        project.access.set_role(user.id, ProjectRole(request.json["role"]))
```
tomasMizera, line 287: Might be good to check if `request.json["role"]` is some valid role (maybe the `ProjectRole` constructor is doing that already) and if not, return error

varmar05: it is a required field of enum type, so error would be caught by openapi validator with 400 error response

varmar05:
> Sign the licence :)

I keep ignoring this. I signed it on master :-) so merge/rebase would do the job.

PR #71, opened 2023-04-28
Project version table updates

MarcelGeo:
- [Added disabling and default count of versions option to project version](https://github.com/MerginMaps/mergin/commit/72634c172fa0d791cb9d56c199474202812fbacf)
- [Added ability do hide pagination
- Added ability to add footer to table

**Next changes**
- Added types
- added versions to store directly, not to project EnhancedObject
- Fix for missing import in ProjectsListView

### DIFF
--- a/web-app/packages/lib/src/modules/project/types.ts
+++ b/web-app/packages/lib/src/modules/project/types.ts
@@ -301,4 +301,6 @@ export interface ProjectVersionsPayload {
   count: number
 }
 
-export type ProjectVersionsItem = ProjectVersion & { disabled: boolean }
+export interface ProjectVersionsItem extends ProjectVersion {
+  disabled: boolean
+}
--- a/web-app/packages/lib/src/modules/project/types.ts
+++ b/web-app/packages/lib/src/modules/project/types.ts
@@ -285,10 +285,20 @@ export interface ProjectTemplate {
 export type EnhancedProjectDetail = ProjectDetail & {
   files: Record<string, ProjectListItemFiles>
   path: string
-  versions?: ProjectVersion[]
 }
 
 export interface UpdateProjectAccessParams {
   user_id: number
   role: ProjectRoleName
 }
+
+export interface FetchProjectVersionsPayload extends ProjectParams {
+  params: FetchProjectVersionsParams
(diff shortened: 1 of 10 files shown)


MarcelGeo:
We must fix missing api for project version.

web-app/packages/lib/src/modules/project/types.ts
```diff
@@ -285,10 +285,20 @@ export interface ProjectTemplate {
...
+export interface FetchProjectVersionsPayload extends ProjectParams {
+  params: FetchProjectVersionsParams
+}
+
+export interface ProjectVersionsPayload {
+  versions: ProjectVersion[]
+  count: number
+}
+
+export type ProjectVersionsItem = ProjectVersion & { disabled: boolean }
```
lavor, line 304: Is there some reason why you prefer type intersection ('&') here over extending interface?

(similarily few lines above for type EnhancedProjectDetail)

web-app/packages/lib/src/modules/project/views/ProjectVersionsViewTemplate.vue
```diff
@@ -0,0 +1,293 @@
...
+    ...mapActions('projectModule', ['fetchProjectVersions']),
+    paginating(options) {
+      this.options = options
+      this.fetchVersions()
+    },
+    fetchPage(number) {
+      this.options.page = number
+      this.fetchVersions()
+    },
+    disabledRow(item: ProjectVersionsItem) {
```
lavor, line 265: I would rename this to something more suitable:
* I would add Style suffix as it defines value for "style" and not value for "row".
* It is disabled only when `item.disabled` is `true `- hence it could be something like `disabledOrEnabledRowStyle`

web-app/packages/lib/src/modules/project/types.ts
MarcelGeo: Sometimes type, sometimes interface :). Disabled is not occuring in ProjectVersion, so its safe solution. But, maybe extends is better here :)

lavor (approved):

varmar05 (approved):

PR #73, opened 2023-05-02
put help function in tests.utils

harminius:
this function does whole push - start, upload, finish for added files

### DIFF
--- a/server/mergin/tests/utils.py
+++ b/server/mergin/tests/utils.py
@@ -193,3 +193,26 @@ def file_info(project_dir, path, chunk_size=1024):
         "mtime": datetime.fromtimestamp(os.path.getmtime(abs_path), tzlocal()),
         "chunks": [str(uuid.uuid4()) for i in range(math.ceil(f_size / chunk_size))],
     }
+
+
+def push_project_file_add(workspace, project, data, client):
+    """ Push added file - start, upload, finish """
+    resp = client.post(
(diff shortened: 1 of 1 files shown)


server/mergin/tests/utils.py
```diff
@@ -193,3 +193,26 @@ def file_info(project_dir, path, chunk_size=1024):
         "mtime": datetime.fromtimestamp(os.path.getmtime(abs_path), tzlocal()),
         "chunks": [str(uuid.uuid4()) for i in range(math.ceil(f_size / chunk_size))],
     }
+
+
+def push_project_file_add(workspace, project, data, client):
```
varmar05, line 198: no need of workspace, it is already in project object

varmar05: maybe rename to push_file_to_project or upload_file_to_project ?

server/mergin/tests/utils.py
```diff
@@ -193,3 +193,26 @@ def file_info(project_dir, path, chunk_size=1024):
...
+    """ Push added file - start, upload, finish """
+    resp = client.post(
+        f"/v1/project/push/{workspace.name}/{project.name}",
+        data=json.dumps(data, cls=DateTimeEncoder).encode("utf-8"),
```
varmar05, line 202: It will be safer to calculate data here, not pass as arg

varmar05 (changes requested):

PR #78, opened 2023-05-23
custom field class that filters whitespaces

harminius:
CU: https://app.clickup.com/t/862jv6teu

GL: https://gitlab.lutraconsulting.co.uk/mergin/mergin/-/merge_requests/812 - apply in workspace create form

rename usernames: https://gitlab.lutraconsulting.co.uk/lutraconsulting/operations/-/merge_requests/5

Change:

when logging in, server removes trailing spaces from username (autocomplete on mobiles added invisible space which lead to login failures)

Custom class for text fields was created and used at any place we want to remove leading and trailing spaces

In following user inputs there is already check for trailing spaces and they do not use forms:
- create workspace invitation
- add project
- clone project

### DIFF
--- a/server/mergin/auth/forms.py
+++ b/server/mergin/auth/forms.py
@@ -56,7 +56,7 @@ class RegisterUserForm(FlaskForm):
         "Username",
         validators=[validators.Length(min=4, max=25), username_validation],
     )
-    email = StringField(
+    email = CustomStringField(
         "Email Address",
         validators=[DataRequired(), Email()],
     )
--- a/server/mergin/auth/forms.py
+++ b/server/mergin/auth/forms.py
@@ -16,11 +16,7 @@ from wtforms.validators import (
 )
 
 from .models import User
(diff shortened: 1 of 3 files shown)


server/mergin/auth/forms.py
```diff
@@ -51,20 +47,18 @@ def __call__(self, form, field):
...
 
 
 class RegisterUserForm(FlaskForm):
-    username = StringField(
+    username = CustomStringField(
         "Username",
         validators=[validators.Length(min=4, max=25), username_validation],
-        filters=(whitespace_filter,),
     )
     email = StringField(
```
varmar05, line 59: should be also CustomStringField

varmar05 (changes requested):

server/mergin/auth/forms.py
harminius: good catch

varmar05 (approved):

PR #91, opened 2023-06-21
Cleanup access requests from Project

MarcelGeo:
**BE**

Cleanup access_requests list from ProjectSchema

**FE**

Change mapping of access requests in project detail from ProjectDetail.access_requests to store accessRequests with count

**Block**

merge after https://github.com/MerginMaps/mergin/pull/57

### DIFF
--- a/server/mergin/sync/schemas.py
+++ b/server/mergin/sync/schemas.py
@@ -172,7 +172,6 @@ class ProjectSchema(ma.SQLAlchemyAutoSchema):
     id = fields.UUID()
     files = fields.Nested(FileInfoSchema(), many=True)
     access = fields.Nested(ProjectAccessSchema())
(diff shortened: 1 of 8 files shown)


server/mergin/sync/schemas.py
```diff
@@ -172,7 +172,6 @@ class ProjectSchema(ma.SQLAlchemyAutoSchema):
     id = fields.UUID()
     files = fields.Nested(FileInfoSchema(), many=True)
     access = fields.Nested(ProjectAccessSchema())
-    access_requests = fields.Nested(ProjectAccessRequestSchema(), many=True)
```
varmar05, line 175: Can you also remove relationship from models so it is not fetched from db with project?

MarcelGeo: cleaned from model removing backref for access_requests

varmar05 (approved):

PR #93, opened 2023-06-23
Add sort in json fields

varmar05:
Add support to sort by keys in db JSON fields in query parameters as `json_field.key`, e.g. `?order_params=meta.storage ASC`

### DIFF
--- a/server/mergin/utils.py
+++ b/server/mergin/utils.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
 
 from collections import namedtuple
+from enum import Enum
 from flask_sqlalchemy import Model
 from sqlalchemy import Column, JSON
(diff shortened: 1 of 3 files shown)


harminius:
looks good!

server/mergin/utils.py
```diff
@@ -1,31 +1,93 @@
...
+OrderParam = namedtuple("OrderParam", "name direction")
+
+
+def split_order_param(order_param: str) -> Optional[OrderParam]:
+    """Split db query order parameter"""
+    try:
+        col, order = order_param.strip().split(" ")
+    except ValueError:
+        return
+    if order.lower() in ["asc", "desc"]:
```
MarcelGeo, line 21: Just small refactor, if possible:

Use an enum for order directions instead of string literals. Here, you can check for members of enum, not hardcoded list.

MarcelGeo (approved):

PR #96, opened 2023-07-11
Complete removal days calculation #862jybpje

harminius:
CU: https://app.clickup.com/t/862jybpje

The period in which user user can create project with the same name as they deleted was increasing over time. This fixes the calculation of the expiration variable that is used in the error message.

Screenshot from 2023-07-11 10-47-44

### DIFF
--- a/server/mergin/sync/public_api_controller.py
+++ b/server/mergin/sync/public_api_controller.py
@@ -12,7 +12,7 @@ from typing import Dict
 from urllib.parse import quote
 import uuid
 from time import time
-from datetime import datetime, timedelta
+from datetime import datetime
 import psycopg2
 from blinker import signal
 from connexion import NoContent, request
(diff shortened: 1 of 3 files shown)


server/mergin/tests/test_project_controller.py
```diff
@@ -2130,6 +2130,12 @@ def test_inactive_project(client, diff_project):
     )
     assert resp.status_code == 409
     assert "Project with the same name is scheduled for deletion" in resp.json["detail"]
+    assert (
```
varmar05, line 2133: merge to single assert for detail field

server/mergin/tests/test_project_controller.py
```diff
@@ -2141,6 +2147,12 @@ def test_inactive_project(client, diff_project):
...
```
varmar05, line 2150: same as above

server/mergin/sync/public_api_controller.py
```diff
@@ -176,7 +176,10 @@ def add_project(namespace):  # noqa: E501
         ).first()
         if proj:
             if proj.removed_at:
-                expiration = format_time_delta(datetime.utcnow() - proj.removed_at)
+                expiration = format_time_delta(
+                    timedelta(days=current_app.config["DELETED_PROJECT_EXPIRATION"])
```
varmar05, line 180: maybe create expiration method / property in Project class?

PR #97, opened 2023-07-18
Consolidate user removal

varmar05:
API for user account removal were updated:
- user can initiate soft delete - object scheduled for removal 
- admin can initiate hard delete - object immediately removed
- admin can temporarily de/activate user account

Private admin related user APIs were renamed `/auth` -> `/admin`.
New attribute was added  to User object response - `scheduled_removal`, e.g.

```
GET /app/admin/user/<username>
{
'active': False, 
'email': 'alice@mergin.com', 
'id': 2, 
'is_admin': False, 
'profile': {...}, 
'scheduled_removal': '2023-07-23T12:47:24Z',
'username': 'alice', 
'verified_email': True
}
```

Other changes:
Endpoints emit signals so other parts of application can listen on them.


.python-version
MarcelGeo, line 1: Hide this file

server/mergin/auth/controller.py
```diff
@@ -423,11 +423,11 @@ def update_user(username):  # pylint: disable=W0613,W0612
 @auth_required(permissions=["admin"])
 def delete_user(username):  # pylint: disable=W0613,W0612
     user = User.query.filter_by(username=username).first_or_404("User not found")
-    user_deleted.send(user)
     inactivate_user(user)
-    db.session.delete(user)
-    db.session.commit()
-    return "", 200
+    user_account_closed.send(user)
```
MarcelGeo, line 427: Do we need emit this event?

varmar05: we do, there is a listener which removes workspace related bits (like references in workspace tables)

MarcelGeo:
# FE updates

- Added proper icons for action buttons in Profile view
- Update endpoints for /app/auth/user*

## Admin

### Active user

image

### Inactive user

image

### Tootlip for "Delete account"

image

### Dialog for deleting account

image

## Webapp

### User profile view

image

MarcelGeo (approved):
Approved and tested BE / FE

PR #99, opened 2023-07-21
Fix failing changeset details / file restore

varmar05:
Occasionally, during preview of file changeset details or downloading file at older version action failed with 'Missing base file' error. Two issues were identified:

- GET /v1/resource/changesets endpoint did not check for existence of basefile; 
 --> FIX: call gpkg restore function if basefile not present (e.g. for old version) 
- in case of gpkg file restoration from 'added' basefile it was found that metadata coming from db query did not contain 'location' key because we do not store it
--> FIX: missing key was added on the fly as it is computed property only cached in db, also test fixture was updated so it would match real scenario and won't hide failure


harminius:
looks good to me

harminius (approved):
looks good to me

wonder-sk (approved):

PR #102, opened 2023-07-26
remove unimportant emails #861n3wzau

harminius:
CU: https://app.clickup.com/t/861n3wzau

Let's not send emails which are not important - remove templates and logic.


server/mergin/sync/public_api_controller.py
```diff
@@ -662,31 +661,6 @@ def update_project(namespace, project_name):  # noqa: E501  # pylint: disable=W0
             privileges.append("upload")
         if user_profile.user.id in project.access.readers:
             privileges.append("download")
-        subject = "Project access modified"
```
varmar05, line 665: remove everything email related, since comment "# send email notifications"

varmar05:
@harminius could you be more specific in description? Like what emails were removed, what signals / handlers, etc.

server/mergin/sync/public_api_controller.py
harminius: thanks

varmar05 (approved):

PR #109, opened 2023-09-05
#861n6f0z0: Username to failed syncs

luxusko:
[#861n6f0z0](https://app.clickup.com/t/861n6f0z0)

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -165,10 +165,10 @@ class Project(db.Model):
 
         return history
 
-    def sync_failed(self, client, error_type, error_details, username):
+    def sync_failed(self, client, error_type, error_details, user_id):
         """Commit failed attempt to sync failure history table"""
         new_failure = SyncFailuresHistory(
-            self, client, error_type, error_details, username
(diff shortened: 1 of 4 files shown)


server/mergin/auth/db_events.py
```diff
@@ -5,10 +5,22 @@
 from .app import force_delete_user
 from .. import db
 from ..auth.models import User
+from mergin.sync.models import SyncFailuresHistory
+
+
+def remove_from_SyncFailuresHistory(user: User):
```
varmar05, line 11: please convert to lower case and add doc string

varmar05 (changes requested):
@luxusko could you update tests?

varmar05:
@tomasMizera is there any reason why we do not wan't to go with user id as FK and cascade delete?
I imagine we only analyse such data using some BI.

tomasMizera:
> @tomasMizera is there any reason why we do not wan't to go with user id as FK and cascade delete? I imagine we only analyse such data using some BI.

Well, it would work with CE, but I am afraid other versions could not use this

PR #121, opened 2023-10-13
Minus from cookie parsing

MarcelGeo:
- it was not possible to remove data from cookies, method for removing failed


harminius (approved):
nice

PR #124, opened 2023-10-19
Add endpoint to rename a project

harminius:
Task: https://github.com/MerginMaps/MerginMaps-Cloud-TEST/issues/1

URL: PATCH /v2/projects/<project_id>
Body: `{ "name": "new_project_name"}`

- check the user have permission to update the project
- check the new name uses allowed characters
- check if a project with the new name already exists

### DIFF
--- a/server/mergin/sync/public_api_v2_controller.py
+++ b/server/mergin/sync/public_api_v2_controller.py
@@ -4,14 +4,17 @@
 
 from datetime import datetime
 
-from connexion import NoContent
-from flask import abort
+from connexion import NoContent, request
+from flask import abort, jsonify, make_response, current_app
 from flask_login import current_user
 
 from mergin import db
 from mergin.auth import auth_required
-from mergin.sync.models import Project
 from mergin.sync.permissions import ProjectPermissions, require_project_by_uuid
+
+from .errors import InvalidProjectName, ProjectNameAlreadyExists
+from .models import Project
+from .utils import is_name_allowed
 
 
 @auth_required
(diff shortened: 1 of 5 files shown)


server/mergin/sync/public_api_v2_controller.py
```diff
@@ -43,3 +44,32 @@ def delete_project_now(id_):
...
+    :param id_: Project_id
+    :type id_: str
+    """
+    project = require_project_by_uuid(id_, ProjectPermissions.Update)
+    new_name = request.json["name"]
+
+    if not is_name_allowed(new_name):
+        return (
+            jsonify(
+                code="InvalidProjectName", detail="Entered project name is invalid"
```
varmar05, line 62: I think we should stick with existing form of returning customized errors (this is where we aimed migrate to)
https://github.com/MerginMaps/mergin/blob/develop/server/mergin/sync/errors.py

server/mergin/tests/test_public_api_v2.py
```diff
@@ -37,3 +37,23 @@ def test_delete_after_schedule(client):
...
+    data = {"name": "new_project_name"}
+    response = client.patch(f"v2/projects/{project.id}", json=data)
+    assert response.status_code == 204
+    assert project.name == "new_project_name"
+    # name already exists
+    response = client.patch(f"v2/projects/{project.id}", json=data)
+    assert response.status_code == 409
+    # invalid project name
+    response = client.patch(f"v2/projects/{project.id}", json={"name": ".new_name"})
+    assert response.status_code == 400
```
varmar05, line 55: 400 is generic error returned by connexion validation, e.g. missing payload param, I suggest assert here our own error code is returned

server/mergin/tests/test_public_api_v2.py
```diff
@@ -37,3 +37,23 @@ def test_delete_after_schedule(client):
...
+    response = client.patch(
+        f"v2/projects/{project.id}", json={"name": " new_project_name"}
+    )
+    assert response.status_code == 400
```
varmar05, line 59: add test for 403?

varmar05: I meant integration test that in this endpoint we use correct check of permissions, e.g. call endpoint as not-owner.

server/mergin/sync/public_api_v2_controller.py
varmar05: ok, let's deal with this in separate PR with errors clean up

varmar05 (approved):

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -43,3 +44,32 @@ def delete_project_now(id_):
...
+            ),
+            400,
+        )
+    new_name_exists = Project.query.filter_by(
+        workspace_id=project.workspace_id, name=new_name
```
harminius, line 67: @luxusko I think we should filter by `name=new_name.strip()`, right? 

because if there is a trailing space in new_name, it is not catched by is_name_allowed and here we should check the name which we want to write to db (there could be already a different project without the whitespace)

I didn't notice you moved the .strip() to the end.

PR #127, opened 2023-10-31
Fix project download

varmar05:
This PR mitigates risk of performance penalty for file download:
- do not load project.files from db if not needed
- in download endpoint use db query to fetch file location instead of filtering in app


mostlyAtNight (approved):
Looks good to me, thank you

PR #132, opened 2023-11-25
Change PWD to "." in docker-compose.yml

alhirzel:
docker-compose struggles with PWD sometimes, especially when the command is not run from a shell (e.g. via systemd or via ansible). This commit uses a more idiomatic relative path in the compose file, fixing these issues.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -52,6 +52,6 @@ services:
       - "8080:80"
     volumes:
       - ./projects:/data  # map data dir to host
-      - ${PWD}/nginx.conf:/etc/nginx/conf.d/default.conf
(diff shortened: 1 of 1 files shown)


tomasMizera:
Hi @alhirzel, thanks for your PR, I tried it out and works. Please note that this `docker-compose` file is not meant for production use and needs to be adjusted by your needs.

PeterPetrik:
Hi @alhirzel can you please add your GitHub name to this list please? https://github.com/MerginMaps/server/blob/master/LICENSES/CLA-signed-list.md

alhirzel:
@PeterPetrik see #146

tomasMizera:
Thanks @alhirzel !

PR #152, opened 2024-01-11
Add build hash variable

harminius:
[Issue](https://github.com/MerginMaps/MerginMaps-Cloud-TEST/issues/144)

In this PR the new `BUILD_HASH` variable is put in `/config`. It gets the correct value during the build, see [GitLab CE part](https://gitlab.lutraconsulting.co.uk/mergin/mergin-ce/-/merge_requests/17)


varmar05 (approved):

varmar05:
@harminius I think you need to add BUILD_HASH to dockerfile (not sure in which PR, this one might be good as well)

PR #175, opened 2024-02-14
Enh app section header

MarcelGeo:
**Details**

:hammer: Updates:
- added header slot to app-section
- perf: do not call /access twice
- zindex opts for  modal (dialog)
- autocomplete hinting on focus

:rose: Design
- smaller font in tables
- avatars bg color
- warning button redesign
- margin: 0 for paragraphs
- min-width set to 0 for selected elements (there are hardcoded values in some browsers, then flex algorithm is not running properly)

### DIFF
--- a/web-app/packages/lib/src/modules/dashboard/components/DashboardAccessRequestsRow.vue
+++ b/web-app/packages/lib/src/modules/dashboard/components/DashboardAccessRequestsRow.vue
@@ -6,54 +6,38 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
 
 <template>
   <app-container
-    v-if="currentNamespace && accessRequests && accessRequestsCount > 0"
+    v-if="
+      userStore.currentWorkspace &&
+      projectStore.accessRequests &&
+      projectStore.accessRequestsCount > 0
+    "
   >
     <app-section>
       <template #title
         >Requests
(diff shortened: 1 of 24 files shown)


web-app/packages/lib/src/modules/dashboard/components/DashboardAccessRequestsRow.vue
```diff
@@ -6,54 +6,38 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
...
-  name: 'DashboardAccessRequestsRow',
-  components: { AppSection, AppContainer },
-  computed: {
-    ...mapState(useProjectStore, [
-      'accessRequests',
-      'accessRequestsCount',
-      'currentNamespace'
-    ])
-  },
-  watch: {
```
varmar05, line 42: on workspace switch (e.g. in EE) do we have reload of access requests?

MarcelGeo: yes. setup is called on every mount of component. I tested it and everything was ok. Maybe @jozef-budac will some use case for such watching of workspace change.

varmar05: ok, thanks

varmar05 (approved):

PR #138, opened 2023-12-11
Redesign projects table

MarcelGeo:
**Description**

Scope: Projects table in /projects path with pagination / searching / sorting

image

:exclamation: merge after https://github.com/MerginMaps/server/pull/136

TODO in next PRs:

- create project dialog
- dialog module redesign

PR #178, opened 2024-02-19
Get /app/project/{project_id}/access

luxusko:
Resolves #138

### DIFF
--- a/server/mergin/sync/private_api.yaml
+++ b/server/mergin/sync/private_api.yaml
@@ -329,16 +329,37 @@ paths:
         "404":
           $ref: "#/components/responses/NotFoundResp"
   /project/{id}/access:
+    parameters:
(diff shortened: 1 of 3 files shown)


MarcelGeo:
@luxusko I don't see tests for new endpoint.

MarcelGeo (changes requested):

harminius (commented):
I agree with MarceGeo.

MarcelGeo:
Logic: @luxusko , agreed with @varmar05 

    in case of GLOBAL_WRITE - everybody with reader 🔽 in project.access will be writer 🔼 , everybody with owner in project.access will be owner (NOT WRITER)

MarcelGeo (approved):
Looks ok now :)

PR #179, opened 2024-02-20
add generated file

harminius:
[Task description](https://github.com/MerginMaps/MerginMaps-Cloud-TEST/issues/138)

### DIFF
--- a/LICENSES/EE-used-libs.json
+++ b/LICENSES/EE-used-libs.json
@@ -1,13 +1,5 @@
 [
     {
-        "library": "aiohttp",
-        "version": "3.9.1"
-    },
-    {
-        "library": "aiosignal",
-        "version": "1.3.1"
-    },
-    {
         "library": "alembic",
         "version": "1.12.0"
     },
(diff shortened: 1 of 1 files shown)


MarcelGeo (changes requested):

LICENSES/EE-used-libs.json
```diff
@@ -0,0 +1,626 @@
...
+    {
+        "library": "colorama",
+        "version": "0.4.5"
+    },
+    {
+        "library": "connexion",
+        "version": "2.14.1"
+    },
+    {
+        "library": "Deprecated",
```
MarcelGeo, line 83: this is library?

MarcelGeo (changes requested):

LICENSES/EE-used-libs.json
harminius: it is also included in pipfile.lock next to other libs

MarcelGeo (approved):

PR #184, opened 2024-02-28
Keep objects referenced in database

varmar05:
Summary:
- stop removing project
- keep resolved access requests
- migrate registration_date column to user table
- use user_id in login history
- stop removing users from db
- use user.id in project.removed_by


varmar05:
@MarcelGeo could you please verify this does not have impact on frontend?

MarcelGeo (approved):
For me, schemas are tansparent with .yml . Approve

PR #188, opened 2024-03-11
Project collaborators

MarcelGeo:
:rose: Design:
- Added collaborators tab

image

:hammer: API

- modify PATCH project access to accept alsow public: boolean

:hammer: 

- implement PATCH /project/{project uuid}/access instead of batch updating of all readernames, in project.access
- add label instead of children on buttons


web-app/packages/lib/src/modules/user/views/ProfileViewTemplate.vue
```diff
@@ -104,57 +104,58 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
...
+            :class="[
+              'flex flex-column align-items-start',
+              'row-gap-2',
+              'md:align-items-center md:flex-row'
+            ]"
+          >
+            <div class="flex-grow-1">
+              <p class="font-semibold my-2">Close account</p>
+              <span class="text-xs opacity-80"
+                >Your account will be closed. In case you are an owner of a
```
varmar05, line 143: This text about workspace ownership should be in EE, here it does not have much sense.

MarcelGeo: discussed with @tomasMizera -> it's ok.

PR #192, opened 2024-03-12
Enhancements accross application

MarcelGeo:
:hammer:  Fixes:

- ui: :iphone:  action buttons for project

image (10)

- Infinite in full storage banner in case of diskusage > storage

- Overflow wrap for email in ProfileView
image

Feature:
- New Avatars handling:
_without Full Name_
image

_with only one name_
image

_with full name_
image

_example for access_
image

### DIFF
--- a/web-app/packages/lib/src/common/components/FullStorageWarningTemplate.vue
+++ b/web-app/packages/lib/src/common/components/FullStorageWarningTemplate.vue
@@ -31,11 +31,13 @@ import { useUserStore } from '@/main'
 const userStore = useUserStore()
 
 const usage = computed(() =>
-  Math.floor(
-    (userStore.currentWorkspace?.disk_usage /
-      userStore.currentWorkspace?.storage) *
-      100
-  )
+  userStore.currentWorkspace?.disk_usage > userStore.currentWorkspace?.storage
+    ? 0
+    : Math.floor(
+        (userStore.currentWorkspace?.disk_usage /
+          userStore.currentWorkspace?.storage) *
+          100
+      )
 )
 
 /** Handle open state with connection to message close button */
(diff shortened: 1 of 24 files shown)


web-app/packages/lib/src/common/components/FullStorageWarningTemplate.vue
```diff
@@ -31,11 +31,13 @@ import { useUserStore } from '@/main'
 const userStore = useUserStore()
 
 const usage = computed(() =>
-  Math.floor(
-    (userStore.currentWorkspace?.disk_usage /
-      userStore.currentWorkspace?.storage) *
-      100
-  )
+  userStore.currentWorkspace?.disk_usage > userStore.currentWorkspace?.storage
+    ? 0
```
varmar05, line 35: should not be 100%?

MarcelGeo: if storage === 0 , this is not possible to make disk_usage / storage. We do not want to show this banner in such edge case.

varmar05: but disk usage can be > storage in other cases, like plan downgrade?

MarcelGeo: yeah, but in these cases -> it is not reasonable to warn user about full storage :)

varmar05: ah, that is only warning, sorry, I missed it completely

varmar05 (approved):

jozef-budac:
Just check it on lasters RC and it looks good to me :)

PR #193, opened 2024-03-13
make revision id sequence continuous

harminius:
Revision ids sequence for CE migration was broken in Dec 2022. After this fix, the sequence is continues again and does not cause confusions.

Screenshot from 2024-03-12 14-34-47

### DIFF
--- a/server/migrations/community/1fcbea2a0f2c_drop_namespace_related_objects.py
+++ b/server/migrations/community/1fcbea2a0f2c_drop_namespace_related_objects.py
@@ -5,7 +5,7 @@
 """Drop namespace related objects
 
 Revision ID: 1fcbea2a0f2c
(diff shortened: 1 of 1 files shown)


varmar05:
it was a deliberate branching point, migration 35af0c8be41e is a head for community users but does not make sense for EE/SaaS where is should never be run

to simplify things I think we can try to remove that single purpose migration since it is quite old (2021.6.1->2023.2.0) and community will be linear again - but that would need to be done with caution (e.g. unstamp from db for already migrated dbs) and with proper documentation.

MarcelGeo:
@harminius @varmar05, Shall we close this PR?

MarcelGeo:
We will fix it in separate ticket.

PR #199, opened 2024-03-19
Allow share project parameter and check GLOBAL_ variables

MarcelGeo:
**Details**

Ticket: https://github.com/MerginMaps/MerginMaps-Cloud-TEST/issues/215

- Make sharing buttons visibility configurable in collaborators tab (allowShare: true props to collaborators)
- If there is one of GLOBAL_* variable true, we will hide share project button.

GLOBAL_READ is enabled:

image

There is no GLOBAL_READ/WRITE/ADMIN

image

:rose: Minor design fixes:

- enh: Laptops - change break point for sidebar overlay to 1200px

### DIFF
--- a/server/application.py
+++ b/server/application.py
@@ -34,6 +34,9 @@ application = create_app(
         "SERVER_TYPE",
         "COLLECT_STATISTICS",
         "USER_SELF_REGISTRATION",
+        "GLOBAL_ADMIN",
+        "GLOBAL_READ",
+        "GLOBAL_WRITE"
     ]
 )
 register_stats(application)
--- a/server/application.py
+++ b/server/application.py
@@ -36,7 +36,7 @@ application = create_app(
         "USER_SELF_REGISTRATION",
         "GLOBAL_ADMIN",
         "GLOBAL_READ",
-        "GLOBAL_WRITE"
+        "GLOBAL_WRITE",
     ]
 )
 register_stats(application)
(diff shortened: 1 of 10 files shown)


server/application.py
```diff
@@ -34,6 +34,9 @@
         "SERVER_TYPE",
         "COLLECT_STATISTICS",
         "USER_SELF_REGISTRATION",
+        "GLOBAL_ADMIN",
```
varmar05, line 37: do you need all of them? (I'd say READ is enough)

web-app/packages/app/src/modules/project/views/ProjectCollaboratorsView.vue
```diff
@@ -8,6 +8,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
   <project-collaborators-view-template
     v-if="isProjectOwner"
     @share="openShareDialog"
+    :allow-share="!instanceStore.globalRolesEnabled"
```
varmar05, line 11: how about GLOBAL_READ=false which means that all users are just guests without any explicit access to anything - in that case you want to allow sharing, no?

varmar05: ah, I see it now in that function - and yet I'd simplify it to check of READ flag, WRITE=1 READ=0 does not make much sense.

server/application.py
MarcelGeo: I should be also GLOBAL_READ=False and GLOBAL_WRITE=True ,  no?

web-app/packages/app/src/modules/project/views/ProjectCollaboratorsView.vue
MarcelGeo: same as above... ?

server/application.py
varmar05: ah yes, you are right

web-app/packages/app/src/modules/project/views/ProjectCollaboratorsView.vue
varmar05: let's keep it as it is, I just do not know why would anyone set WRITE true but READ false

varmar05 (approved):

PR #200, opened 2024-03-19
add combined permission - ws admin or project owner

harminius:
[Task](https://github.com/MerginMaps/MerginMaps-Cloud-TEST/issues/215)
[EE part](https://gitlab.lutraconsulting.co.uk/mergin/mergin/-/merge_requests/1012/)

Permission check for project admin access.

### DIFF
--- a/server/mergin/sync/permissions.py
+++ b/server/mergin/sync/permissions.py
@@ -10,7 +10,7 @@ from flask_login import current_user
 
 from .utils import is_valid_uuid
 from ..auth.models import User
-from .models import ProjectAccess, Project, Upload
+from .models import ProjectAccess, Project, Upload, ProjectRole
 
 
 def _is_superuser(f):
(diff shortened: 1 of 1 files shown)


server/mergin/sync/permissions.py
```diff
@@ -239,3 +239,27 @@ def get_user_project_role(project: Project, user: User) -> Optional[str]:
...
+    :rtype bool
+    """
+    if user.is_anonymous or not user.active:
+        return False
+    workspace = current_app.ws_handler.get(workspace_id)
+    if not workspace:
+        return False
+    if workspace.user_has_permissions(current_user, "admin"):
+        return True
+    projects = Project.query.filter_by(workspace_id=workspace.id).all()
```
varmar05, line 261: can we check this within single query instead of getting all projects just to check permissions on python side?

server/mergin/sync/permissions.py
```diff
@@ -239,3 +239,27 @@ def get_user_project_role(project: Project, user: User) -> Optional[str]:
         return "writer"
     if ProjectPermissions.Read.check(project, user):
         return "reader"
+
+
+def project_admin_permissions(workspace_id, user: User) -> bool:
```
varmar05, line 244: I wonder if we need this in CE at all since this is kind of workspace permission only for a particular use case in EE

varmar05:
logic moved to EE codebase

PR #208, opened 2024-04-05
Store last seen workspaces in local storage

MarcelGeo:
**Details**

Reaction to ticket https://github.com/MerginMaps/server-private/issues/2229

:hammer: 
 Store last seen workspace(s) in local storage instead of cookies. This PR is mostly only refactor, but following improvements are coming:
 
 - local storage is more robust than cookies
 - currentWorkspace cookies will not be sending in every request on API :1st_place_medal: 
 - there is not need to handle expiring of cookies
 - last workspace will be array instead key value (used for showing last seen workspace to user)
 
:heavy_plus_sign: New feature:

- We can now use Proxy based storing to local storage with ability to set value and sync it with localStorage. In case of array value in Proxy, use spread operator (push/pop ... are not reactive). This is experimental feature.

```javascript
const storage = createStorage('name-of-key', 0)

storage.value = 1
```

### DIFF
--- a/web-app/packages/lib/src/modules/user/store.ts
+++ b/web-app/packages/lib/src/modules/user/store.ts
@@ -386,7 +386,7 @@ export const useUserStore = defineStore('userModule', {
         return
       }
       this.setWorkspaceId({ id: payload.workspaceId })
-      if (!payload.skipSavingInCookies) {
+      if (!payload.skipStorage) {
         // Append current workspace with last seen parameter
         this.lastWorkspaces.value = [
           ...this.lastWorkspaces.value.filter(
@@ -426,18 +426,18 @@ export const useUserStore = defineStore('userModule', {
     /**
      * Checks the current workspace that the user is in.
      *
-     * Tries to get the current workspace ID from the cookies.
+     * Tries to get the current workspace ID from the local storage array of last seen ws.
      * If not found, tries to use the preferred workspace ID from the user profile.
      * If still not found, defaults to using the first workspace in the user's list of workspaces.
      *
-     * Sets the current workspace in the store, and saves to cookies if needed.
+     * Sets the current workspace in the store, and saves to local storage if needed.
      */
     async checkCurrentWorkspace() {
       try {
         await this.getWorkspaces()
         let foundWorkspace = this.getLastWorkspaceFromStorage()
         if (this.workspaces.length > 0) {
-          let skipSavingInCookies = true
+          let skipStorage = true
           if (!foundWorkspace) {
             // try to use preferred_workspace from user profile response
             const preferredWorkspaceId = this.loggedUser?.preferred_workspace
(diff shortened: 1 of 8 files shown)


web-app/packages/lib/src/modules/user/store.ts
```diff
@@ -380,167 +381,86 @@ export const useUserStore = defineStore('userModule', {
...
+    async cleanupLastSeenWorkspaces() {
+      this.lastWorkspaces.value = this.lastWorkspaces.value.filter(
+        (item) => item.userId !== this.loggedUser?.id
+      )
     },
 
+    /**
+     * Checks the current workspace that the user is in.
+     *
+     * Tries to get the current workspace ID from the cookies.
```
varmar05, line 429: we do not use cookies anymore, please update

web-app/packages/lib/src/modules/user/store.ts
```diff
@@ -380,167 +381,86 @@ export const useUserStore = defineStore('userModule', {
...
-                // do not skip saving in cookies when using preferredWorkspace
-                skipSavingInCookies = false
-              }
+            const preferredWorkspaceId = this.loggedUser?.preferred_workspace
+            foundWorkspace = this.getWorkspaceById({
+              id: preferredWorkspaceId
+            })
+            if (foundWorkspace?.id) {
+              // do not skip saving in cookies when using preferredWorkspace
+              skipSavingInCookies = false
```
varmar05, line 449: do we need this?

varmar05 (commented):
Nice one, looking good.

MarcelGeo:
@varmar05 Just wondering, do we need this function in community edition with one workspace?

web-app/packages/lib/src/modules/user/store.ts
MarcelGeo: I can rename variables . some workspace changes might not be persisted in storage due to specific logic.

varmar05:
> @varmar05 Just wondering, do we need this function in community edition with one workspace?

We don't, if you can port it easily to EE/cloud, it would be good. Nevertheless backend APIs are available also for CE.

web-app/packages/lib/src/modules/user/store.ts
varmar05: cool, just rename in that case

MarcelGeo:
> > @varmar05 Just wondering, do we need this function in community edition with one workspace?
> 
> We don't, if you can port it easily to EE/cloud, it would be good. Nevertheless backend APIs are available also for CE.

We need to investigate potential solutions for porting specific store functions across different environments. It should be useful for error handling of specific cases in EE/Cloud. ```setWorkspace``` should be also ported with with option for storing of ws. I think this is not scope of this PR.

varmar05 (approved):

PR #210, opened 2024-04-12
Collaborators management based on global variables

MarcelGeo:
Resolves #2230 in private repo

Scope: Make logic with GLOBAL_READ, GLOBAL_WRITE and GLOBAL_ADMIN variables to collaborators tab.

image

Enh: :hammer: 

- allow disable items in app-dropdown
- allow choosing role that makes sense
- disable removing collaborators in case of GLOBAL_* variables are enabled
- remove some project.creator checks

Design :rose: 

- add box-shadow for menus and drop downs copied from design system figma
- add informative tag

:question: 

- @varmar05  do we want to update some endpoints logic?

### DIFF
--- a/web-app/packages/lib/src/modules/instance/store.ts
+++ b/web-app/packages/lib/src/modules/instance/store.ts
@@ -12,6 +12,7 @@ import {
 } from '@/modules/instance/types'
 import { useNotificationStore } from '@/modules/notification/store'
 import { useUserStore } from '@/modules/user/store'
+import { GlobalRole, isAtLeastGlobalRole } from '@/common/permission_utils'
 
 export interface InstanceState {
   initData: InitResponse
@@ -30,14 +31,22 @@ export const useInstanceStore = defineStore('instanceModule', {
 
   getters: {
     /**
-     * Checks if global roles are enabled based on the config data.
-     * Returns true if any of the global role flags are truthy, false otherwise.
+     * Retrieves the current global role from the instance configuration data.
+     *
+     * The global role is determined by checking the `global_read`, `global_write`, and `global_admin` flags in the configuration data.
+     * If none of the flags are set, `undefined` is returned, otherwise the index of the first set flag is returned as the global role.
+     *
+     * @param state - The current instance state.
+     * @returns The current global role, or `undefined` if no global role is set.
      */
-    globalRolesEnabled(state): boolean {
+    currentGlobalRole(state): GlobalRole | undefined {
       // eslint-disable-next-line camelcase
       const { global_read, global_write, global_admin } = state.configData
       // eslint-disable-next-line camelcase
-      return [global_read, global_write, global_admin].some((r) => !!r)
+      const foundIndex = [global_read, global_write, global_admin].findIndex(
+        (r) => !!r
+      )
+      return foundIndex < 0 ? undefined : foundIndex
     }
   },
 
--- a/web-app/packages/lib/src/modules/instance/store.ts
+++ b/web-app/packages/lib/src/modules/instance/store.ts
@@ -4,6 +4,7 @@
 
 import { defineStore } from 'pinia'
 
+import { GlobalRole } from '@/common/permission_utils'
 import { InstanceApi } from '@/modules/instance/instanceApi'
 import {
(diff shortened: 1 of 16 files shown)


web-app/packages/lib/src/modules/instance/store.ts
```diff
@@ -30,14 +31,22 @@ export const useInstanceStore = defineStore('instanceModule', {
...
-    globalRolesEnabled(state): boolean {
+    currentGlobalRole(state): GlobalRole | undefined {
       // eslint-disable-next-line camelcase
       const { global_read, global_write, global_admin } = state.configData
       // eslint-disable-next-line camelcase
-      return [global_read, global_write, global_admin].some((r) => !!r)
+      const foundIndex = [global_read, global_write, global_admin].findIndex(
+        (r) => !!r
+      )
+      return foundIndex < 0 ? undefined : foundIndex
```
varmar05, line 49: maybe we could have used -1 for undefined (guest) role

web-app/packages/lib/src/modules/instance/store.ts
```diff
@@ -30,14 +31,22 @@ export const useInstanceStore = defineStore('instanceModule', {
...
+     * Retrieves the current global role from the instance configuration data.
+     *
+     * The global role is determined by checking the `global_read`, `global_write`, and `global_admin` flags in the configuration data.
+     * If none of the flags are set, `undefined` is returned, otherwise the index of the first set flag is returned as the global role.
+     *
+     * @param state - The current instance state.
+     * @returns The current global role, or `undefined` if no global role is set.
      */
-    globalRolesEnabled(state): boolean {
+    currentGlobalRole(state): GlobalRole | undefined {
```
varmar05, line 42: Just a note: does it matter that in EE and cloud there is not such thing? What would happen if backend returned such flag?

web-app/packages/lib/src/modules/project/components/ProjectMembersTable.vue
```diff
@@ -186,6 +196,10 @@ const searchedItems = computed(() =>
   })
 )
 
+function canRemoveMember(item: ProjectAccessDetail) {
```
varmar05, line 199: How is 'leave a project' handled now? Can you leave project anytime regardless of other members, e.g. you are the last member or project.creator.

varmar05:
We probably do not need to worry about backend changes. Even if you assigned a project role that does not make sense (e.g. lower than global role) it should not matter.

But I'd like to simplify that global role, to have only single variable with enum on backend. But that is probably out of the scope of this PR.

MarcelGeo:
> We probably do not need to worry about backend changes. Even if you assigned a project role that does not make sense (e.g. lower than global role) it should not matter.
> 
> But I'd like to simplify that global role, to have only single variable with enum on backend. But that is probably out of the scope of this PR.

Yeah, It would be nice to have GLOBAL_ROLE variable with ProjectRoleName values :)

web-app/packages/lib/src/modules/instance/store.ts
MarcelGeo: Components in other environments are not connected to this variable.

web-app/packages/lib/src/modules/project/components/ProjectMembersTable.vue
MarcelGeo: If you are guest, you can leave project.

```ProjectViewTemplate.vue#251```

web-app/packages/lib/src/modules/instance/store.ts
MarcelGeo: in global variables is not guest. Or?

varmar05: it is not, I mean if all globals are false, we could jsut use -1 (no need to do +1 in code above)

web-app/packages/lib/src/modules/project/components/ProjectMembersTable.vue
varmar05: In CE if everyone is a guest (no global permission is set) I can leave project anytime, right? Even if I created it.

PR #218, opened 2024-04-30
Refactor project access for ee code injection

luxusko:
Resolves https://github.com/MerginMaps/server-private/issues/2218

### DIFF
--- a/server/mergin/sync/private_api.yaml
+++ b/server/mergin/sync/private_api.yaml
@@ -573,7 +573,7 @@ components:
           format: email
           example: john.doe@example.com
         role:
-          description: Workspace (EE) or global (CE) role
+          description: Workspace role
           type: string
           enum:
             - owner
--- a/server/mergin/sync/private_api.yaml
+++ b/server/mergin/sync/private_api.yaml
@@ -572,7 +572,7 @@ components:
           format: email
           example: john.doe@example.com
         role:
-          description: Workspace role, only in EE
+          description: Workspace (EE) or global (CE) role
           type: string
           enum:
             - owner
@@ -604,8 +604,6 @@ components:
               type: string
               format: date-time
               example: 2018-11-30T08:47:58.636074Z
-            projects:
-              $ref: '#/components/schemas/InvitationProjects'
     InvitationProjects:
       nullable: true
       type: object
--- a/server/mergin/sync/private_api.yaml
+++ b/server/mergin/sync/private_api.yaml
@@ -554,25 +554,38 @@ components:
         - id
         - type
         - email
-        - username
         - project_permission
       properties:
         id:
-          description: User id
-          type: integer
+          description: User/Invitation (uu)id
+          oneOf:
+            - type: integer
+            - type: string
           example: 1
         type:
           type: string
           enum:
             - member
+            - invitation
         email:
           type: string
           format: email
           example: john.doe@example.com
+        role:
+          description: Workspace role, only in EE
+          type: string
+          enum:
+            - owner
+            - admin
+            - writer
+            - reader
+            - guest
         username:
+          description: Present only for type `member`
           type: string
           example: john.doe
         name:
+          description: Present only for type `member`
           type: string
           example: John Doe
         project_permission:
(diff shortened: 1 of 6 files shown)


server/mergin/sync/private_api.yaml
```diff
@@ -581,6 +594,46 @@ components:
...
+          description: Present only for type `invitation`
+          type: object
+          required:
+            - "expires_at"
+          properties:
+            expires_at:
+              type: string
+              format: date-time
+              example: 2018-11-30T08:47:58.636074Z
+            projects:
```
varmar05, line 607: Access detail for project contains invitation with information about other projects? What for?

server/mergin/sync/workspace.py
```diff
@@ -281,3 +281,55 @@ def update_project_members(
...
         """Project access base query"""
         return AccessRequest.query.join(Project)
+
+    @staticmethod
+    def project_access(project: Project) -> List[Dict]:
+        """
+        Project access users overview
+        """
+        global_role = None
+        accesses = (
```
varmar05, line 291: I'd simplify a logic bit here: e.g. global_users: all or none and direct users from project.access. For global users set role from GLOBAL flag, for direct users we should have a helper already to get project role (combined project and workspace role).

server/mergin/sync/workspace.py
```diff
@@ -281,3 +281,55 @@ def update_project_members(
...
+        elif Configuration.GLOBAL_READ:
+            global_role = "reader"
+            accesses = accesses[:2]
+        result = []
+        processed_ids = set()
+        for user_ids, role in accesses:
+            for user_id in user_ids:
+                if user_id not in processed_ids:
+                    user = User.query.get(user_id)
+                    result.append(
```
varmar05, line 311: change to dataclass (in line with EE)

server/mergin/sync/workspace.py
```diff
@@ -281,3 +281,36 @@ def update_project_members(
...
+        for member in members:
+            member_access = MemberAccess(
+                id=member.id,
+                username=member.username,
+                role=ws.get_user_role(member),
+                name=member.profile.name(),
+                email=member.email,
+            ).to_dict()
+            # add correct project permission by comparing ws role and project access permission
+            user = User.query.get(member.id)
```
varmar05, line 311: in case of global members this would cause lot of calls to db

I would probably split lists to global_members and project_members and do some magic there:
- if global_members then filter from them project_members and use it for role override
- else filter only project_members from db and use its role directly

varmar05 (changes requested):

server/mergin/sync/workspace.py
harminius: maybe this query is not needed (I just wanted to be safe) 
Tests are happy when I pass `member` object to `get_user_project_role()`

server/mergin/sync/schemas.py
```diff
@@ -330,3 +329,21 @@ def _user_role(self, obj):
         if not self.context.get("user"):
             return
         return obj.get_user_role(self.context.get("user"))
+
+
+class StrInt(fields.Field):
```
varmar05, line 334: how about `StrOrInt` + write docstring so it is clear what it is doing

server/mergin/sync/schemas.py
```diff
@@ -330,3 +329,21 @@ def _user_role(self, obj):
...
+    def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, str) or isinstance(value, int):
+            return value
+        else:
+            raise ValidationError("Field should be str or list")
```
varmar05, line 339: ... or int

server/mergin/sync/schemas.py
```diff
@@ -330,3 +329,21 @@ def _user_role(self, obj):
...
+
+
+class ProjectAccessSchema(ma.SQLAlchemyAutoSchema):
```
varmar05: do we need to inherit SQLAlchemy schema?

varmar05 (approved):

PR #220, opened 2024-05-02
Fix workspace anonymization ce

harminius:
fixes: https://github.com/MerginMaps/server-private/issues/2271

calling .delete() method on already deleted project will not fail but do nothing


varmar05 (approved):

varmar05:
@harminius could you merge master to develop first? Also could you have a look on coveralls issue? Thanks.

PR #229, opened 2024-05-13
Editor permission

MarcelGeo:
Resolves https://github.com/MerginMaps/server-private/issues/2301

## Backend details

Added :heavy_plus_sign: 

- introduce project_handler abstract class with patch to Flask (```current_app.project_handler```) into application
- handle push permission in project_handler
- Allow upload for editors (```Project.permissions.upload``` will check editors array and workspace role, not writers) - https://github.com/MerginMaps/server/pull/229/commits/f8a0e860b6e00e6342fc03603dbab7a659c524e1) - @varmar05 - we can extract this function to project handler and use it in models to check permissions, hm? 
    - In CE server - there is inconsistency between upload permission and checks in endpoints (/project/push will be allowed only for writers) - I think It is ok 

Migration database :department_store: 
- All writers will be added to editors array

### Compatibility issues with other clients

**Python api client**
* operation ```share_remove``` is not removing users from project.access (TODO: :exclamation: Fix it on client side - discussed with @wonder-sk ) 
* operation ```share_add``` could be unstable also

## Frontend

- handle files upload / removing permission with project writer role instead of Project.permissions.upload. We are allowing upload also for editors, FE can not be mapped on this attribute -> **Editor in frontend will be Reader**.
- extended enums and operations on Project permissions
- filter out editor permission in selected applications
- store permission in projects store

:exclamation:  TODO:

- check admin interface in separate PR
- update python-client in separate PR
- merge after https://github.com/MerginMaps/server/pull/218
- ...


server/mergin/sync/models.py
```diff
@@ -639,10 +649,14 @@ def accept(self, permissions):
         """Accept project access request"""
         project_access = self.project.access
         readers = project_access.readers.copy()
+        editors = project_access.editors.copy()
         writers = project_access.writers.copy()
         owners = project_access.owners.copy()
         readers.append(self.requested_by)
         project_access.readers = readers
+        if permissions in ("edit", "write", "owner"):
```
varmar05, line 657: I'd go with `ProjectAccess.set_role`

server/mergin/sync/permissions.py
```diff
@@ -85,6 +85,18 @@ def query(cls, user, as_admin=True, public=True):
...
             )
 
+    class Edit(Base):
+        @classmethod
+        @_is_superuser
+        def check(cls, project, user):
+            return super().check(project, user) and (
+                (
+                    user.id
+                    in project.access.readers
```
varmar05, line 95: editors?

server/mergin/sync/private_api.yaml
```diff
@@ -128,7 +128,9 @@ paths:
                   enum:
                     - owner
                     - write
+                    - edit
                     - read
+                    - editor
```
varmar05, line 133: why edit and editor?

server/mergin/sync/workspace.py
```diff
@@ -65,14 +65,16 @@ def disk_usage(self):
...
 
+    # EDITOR changes for v2, project ignored in CE
     def user_has_permissions(self, user, permissions):
         role = self.get_user_role(user)
         # mergin super-user has all permissions
         if role == "owner":
             return True
 
-        if permissions == "read":
+        if permissions in ("read", "edit"):
```
varmar05, line 75: the rest of code assumes editor being independent of reader

server/mergin/sync/workspace.py
```diff
@@ -83,6 +85,7 @@ def user_has_permissions(self, user, permissions):
     def user_is_member(self, user):
         return True
 
+    # EDITOR add global var and return role?
```
varmar05, line 88: we probably do not even need to set it on workspace level

server/migrations/community/54c2c49fe2c7_add_editors.py
```diff
@@ -0,0 +1,54 @@
+"""add editors
```
varmar05, line 1: add project editors

server/mergin/sync/permissions.py
```diff
@@ -129,6 +141,20 @@ def check(cls, project, user):
                 )
             )
 
+    @classmethod
+    def get_user_project_role(self, project: Project, user: User) -> Optional[str]:
```
varmar05, line 145: How about returning ProjectRole, we would not need to compare strings at other places (e.g. workspace.py)

varmar05 (commented):
Looking good!

server/mergin/sync/private_api.yaml
MarcelGeo: Just typ

server/mergin/sync/models.py
MarcelGeo: ok?

```
def accept(self, permissions):
        """Accept project access request"""
        project_access = self.project.access
        PERMISSION_PROJECT_ROLE = {
            "read": ProjectRole.READER,
            "edit": ProjectRole.EDITOR,
            "write": ProjectRole.WRITER,
            "owner": ProjectRole.OWNER,
        }

        project_access.set_role(
            self.requested_by, PERMISSION_PROJECT_ROLE.get(permissions)
        )
        self.resolve(RequestStatus.ACCEPTED, current_user.id)
        db.session.commit()
 ```

File: server/mergin/sync/schemas.py [hunk 8]
```diff
@@ -57,7 +60,7 @@ def insert_usernames(self, data, **kwargs):
 
 def project_user_permissions(project):
     return {
-        "upload": ProjectPermissions.Upload.check(project, current_user),
+        "upload": ProjectPermissions.Edit.check(project, current_user),
```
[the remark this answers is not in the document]
MarcelGeo: This mapping is used by mobile and mergin client to check if it is possible to make push to server. We can rename it in future upload -> Edit and add new Write key.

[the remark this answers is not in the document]
MarcelGeo: As I mentioned in PR description, there is inconsistency between "upload" and push permissions in case, that user is editor and do something not in editors rule defined (conflit, adding qgs file, ...).

server/mergin/sync/schemas.py
```diff
@@ -138,7 +141,10 @@ class ProjectSchemaForVersion(ma.SQLAlchemyAutoSchema):
     role = fields.Method("_role")
 
     def _role(self, obj):
-        return get_user_project_role(obj.project, current_user)
+        role = ProjectPermissions.get_user_project_role(obj.project, current_user)
+        if not role:
```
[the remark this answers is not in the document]
MarcelGeo: This method returns enum.value ... I am not sure If ```role = fields.Method("_role")``` will transform enum properly to string.

server/mergin/sync/schemas.py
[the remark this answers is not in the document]
varmar05: it is just good to have in in code so we don't forget :-)

server/mergin/sync/schemas.py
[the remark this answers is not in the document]
varmar05: ah ok, we can then simplify to `if role then return role.value` I was just confused with that if none return none

varmar05 (approved):

PR #231, opened 2024-05-20
case insensitive login

harminius:
Make login case insensitive

authenticate() method is used in login through app as well as public login

migration tested on dev clone and went smoothly

### DIFF
--- a/server/mergin/auth/models.py
+++ b/server/mergin/auth/models.py
@@ -33,8 +33,8 @@ class User(db.Model):
     )
 
     __table_args__ = (
-        db.Index("ix_username_insensitive_unique", func.lower(username), unique=True),
-        db.Index("ix_email_insensitive_unique", func.lower(email), unique=True),
+        db.Index("ix_user_username", func.lower(username), unique=True),
+        db.Index("ix_user_email", func.lower(email), unique=True),
     )
 
     def __init__(self, username, email, passwd, is_admin=False):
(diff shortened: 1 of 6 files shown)


server/mergin/auth/models.py
```diff
@@ -32,6 +32,11 @@ class User(db.Model):
         default=datetime.datetime.utcnow,
     )
 
+    __table_args__ = (
+        db.Index("ix_username_insensitive_unique", func.lower(username), unique=True),
```
varmar05, line 36: we can probably remove original unique indexes?

server/migrations/community/07f2185e2428_case_insensitive_unique_idx.py
```diff
@@ -25,9 +25,15 @@ def upgrade():
     conn.execute(
         "CREATE UNIQUE INDEX ix_email_insensitive_unique ON public.user (LOWER(email));"
     )
+    conn.execute("ALTER TABLE public.user DROP CONSTRAINT uq_user_email;")
```
varmar05, line 28: we should follow the same naming convention with new constraints

varmar05 (approved):

server/migrations/community/07f2185e2428_case_insensitive_unique_idx.py
```diff
@@ -0,0 +1,37 @@
...
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "07f2185e2428"
+down_revision = "0e3fc92aeaaa"
```
MarcelGeo, line 15: Check if ook

harminius: updated, thanks

PR #234, opened 2024-06-04
Merge web and gateway images; split up server and celery; enable docker to be run as non-root

domfre:
Merge web and gateway images; split up server and celery; enable docker to be run as non-root


server/Dockerfile
```diff
@@ -19,29 +22,36 @@ RUN apt-get update -y && \
...
+COPY ./Pipfile /app
 
 RUN pip3 install pipenv==2022.7.24
 # for locale check this http://click.pocoo.org/5/python3/
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
 RUN pipenv install --system --deploy --verbose
-RUN pip3 install flower==0.9.7
+RUN pip3 install celery==5.2.2 kombu==5.2.2
```
MarcelGeo, line 38: You can remove flower dependency, but celery is already included in Pipfile

server/Dockerfile
```diff
@@ -19,29 +22,36 @@ RUN apt-get update -y && \
...
+COPY --chown=mergin:mergin . /home/mergin/app
+RUN chmod -R 755 /home/mergin/app
+
+# Set the non-root user as the default user
 USER mergin
 
-COPY ./entrypoint.sh .
-ENTRYPOINT ["./entrypoint.sh"]
+# Set working directory for non root user
+WORKDIR /home/mergin/app
```
MarcelGeo, line 55: I'm not sure for this change. We can leave it in previous directory ```/app```.

server/config.py
```diff
@@ -37,8 +37,6 @@
     )
     sys.exit(1)
 
-bind = "0.0.0.0:5000"
```
MarcelGeo, line 40: We can leave it there. It's just default value.

MarcelGeo (changes requested):
Hi.
- Celery split is ok in docker-compose 
- We can move other changes to development docker-compose instead of official docker compose for deployment. 
- we want to use gateway instead of have one nginx proxy merged with frontend image.

server/config.py
danduk82: We actually removed this on purpose, so the user can define it at runtime. In some use-cases this has practical advantage, and indeed, as this is the default value I don't see why to hardcode it in the config.

server/Dockerfile
danduk82: there were some problems with ownership but after review with @domfre we will try to revert it accordingly

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ networks:
 
 services:
   db:
-    image: postgres:14
+    image: postgis/postgis:14-3.4
```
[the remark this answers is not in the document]
danduk82: @domfre what was the reason for this?  I don't remember what the issue was, can double check this?

docker-compose.yml
```diff
@@ -20,38 +20,75 @@ services:
     restart: always
     networks:
       - merginmaps
-  server:
-    image: lutraconsulting/merginmaps-backend:2024.2.2
```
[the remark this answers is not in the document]
danduk82: a `docker-compose.dev.yml` that includes the build context, very good idea

server/Dockerfile
danduk82: good point, indeed just remove the line and put that in pipfile

server/Dockerfile
domfre: The point is, on the very top you set the WORKDIR to /app, however if we want to make the container to be runnable as non-root, we cannot set the home directory of the non-root user to be the same as the top workdir, otherwise we run into permission errors. So either we rename the top dir to something else or we rename the home directory of the mergin user (i.e. the non-root user).

server/Dockerfile
domfre: Agree, we can omit this

docker-compose.yml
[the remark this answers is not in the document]
domfre: Yes, I agree, we don't need postgis.

danduk82 (approved):
@domfre this looks fine from my side. @MarcelGeo this should implement your requested changes, could you review it again please?

MarcelGeo:
> @domfre this looks fine from my side. @MarcelGeo this should implement your requested changes, could you review it again please?

Thanks @danduk82 and @domfre ... Let me wait for @varmar05 to review nginx web and gateway ... rest of your changes looks brilliant.

docker-compose.yml
```diff
@@ -20,38 +20,75 @@ services:
...
     container_name: merginmaps-web
     restart: always
     depends_on:
-      - server
+      - server-gunicorn
     links:
       - db
-    networks:
-      - merginmaps
-  proxy:
```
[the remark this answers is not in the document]
varmar05: I think it depends on deployment type, I know there are some users already using host nginx as both, proxy and for TLS termination. Also there are users using dockerized proxy as stated in current docker-compose example, so I am bit worried about such changes for inexperienced users.

I agree it could be merged and you guys can certainly do that for your k8s deployments but I would prefer here in public repo to keep is as is even if not optimal. We can instead maybe provide some optimization hints instead?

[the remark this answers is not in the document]
MarcelGeo: I agree with @varmar05  ... It should be useful to have volume for web-app (a.k.a dashboard) nginx config and upgrade It in custom deployment.

danduk82 (approved):
LGTM

MarcelGeo:
> LGTM

There's one unresolved discussion about nginx gateway... could we improve It?

MarcelGeo (approved):

PR #235, opened 2024-06-04
Project access request link to dashboard

harminius:
Fixes https://github.com/MerginMaps/server-private/issues/2254

Link in email leads to dashboard instead project settings.

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -365,7 +365,7 @@ class ProjectAccessDetail:
     role: str
     username: str
     name: Optional[str]
-    project_permission: str
+    project_permission: ProjectRole
     type: str
 
 
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -365,7 +365,7 @@ class ProjectAccessDetail:
     role: str
     username: str
     name: Optional[str]
-    project_permission: ProjectRole
+    project_permission: str
     type: str
 
 
--- a/server/mergin/sync/private_api_controller.py
+++ b/server/mergin/sync/private_api_controller.py
@@ -69,7 +69,7 @@ def create_project_access_request(namespace, project_name):  # noqa: E501
             "html": render_template(
                 "email/project_access_request.html",
                 expire=access_request.expire,
-                link=f"{request.url_root}",
+                link=f"{request.url_root.rstrip('/')}/projects/{namespace}/{project.name}/collaborators",
                 user=current_user.username,
                 username=owner.username,
                 project_name=f"{namespace}/{project.name}",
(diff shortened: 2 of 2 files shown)


server/mergin/sync/private_api_controller.py
```diff
@@ -69,7 +69,7 @@ def create_project_access_request(namespace, project_name):  # noqa: E501
             "html": render_template(
                 "email/project_access_request.html",
                 expire=access_request.expire,
-                link=f"{request.url_root.rstrip('/')}/projects/{namespace}/{project.name}/settings",
+                link=f"{request.url_root}",
```
varmar05, line 72: This would lead to current selected workspace of owner not a project workspace, no?

server/mergin/sync/models.py
```diff
@@ -365,7 +365,7 @@ class ProjectAccessDetail:
     role: str
     username: str
     name: Optional[str]
-    project_permission: str
+    project_permission: ProjectRole
```
varmar05, line 368: do we need this?

server/mergin/sync/private_api_controller.py
harminius: yes, currently there is no way how to pass workspace in url for dashboard page, there is probably same issue for other email links

there is an issue for this: https://github.com/MerginMaps/server-private/issues/2221

server/mergin/sync/models.py
harminius: i thought it is more descriptive than general string, do you want to put it back?

varmar05: I'd stick to ProjectRole, better than using random string, just wondering if something did not break in code - could you check usage?

server/mergin/sync/private_api_controller.py
MarcelGeo: Just change settings -> collaborators page. There is new place for access requests for admins. We do not want to redirect to dashboard. ?workspace_id is working just for subscription page.

MarcelGeo (changes requested):

server/mergin/sync/private_api_controller.py
MarcelGeo: https://app.dev.merginmaps.com/projects/problem/test/collaborators

server/mergin/sync/models.py
MarcelGeo: It's not ProjectRole.... we don't have enum for ProjectPermission (read/write ... )

MarcelGeo (approved):

server/mergin/sync/private_api_controller.py
```diff
@@ -59,7 +59,7 @@ def create_project_access_request(namespace, project_name):  # noqa: E501
     # notify project owners
     owners = (
         User.query.join(UserProfile)
-        .filter(User.verified_email, User.id.in_(project.access.owners))
```
MarcelGeo, line 62: Just wondering do we want to enable it for not verified emails?

PR #239, opened 2024-06-18
Fix: Prevent download of inactive ws zip (without 500)

MarcelGeo:
Fixes https://github.com/MerginMaps/server-private/issues/2381

- prevent 500 error code for case where user is_anonymous and stayed on project tree page and trying to donwload file of inactive ws
- admin is still available to download such project

Error message now (workspace is not active, does not exists for users anymore):
image

### DIFF
--- a/server/mergin/sync/permissions.py
+++ b/server/mergin/sync/permissions.py
@@ -160,22 +160,22 @@ class ProjectPermissions:
 
 def is_active_workspace(workspace):
     """
-    Checks if the given workspace is active and if the current user is an admin or anonymous.
+    Checks if the given workspace is active or if the current user is an admin.
 
     Args:
         workspace (Workspace): The workspace to check.
 
     Returns:
-        bool: True if the workspace is not active and the current user is not an admin or is anonymous, False otherwise.
+        bool: True if the workspace is active or the current user is an admin, False otherwise.
     """
-    if not workspace:
-        abort(404)
     is_admin = not current_user.is_anonymous and current_user.is_admin
     return workspace.is_active or is_admin
 
 
 def require_project(ws, project_name, permission):
     workspace = current_app.ws_handler.get_by_name(ws)
+    if not workspace:
+        abort(404)
     if not is_active_workspace(workspace):
         abort(404, "Workspace doesn't exist")
     project = (
@@ -200,6 +200,8 @@ def require_project_by_uuid(uuid: str, permission: ProjectPermissions, scheduled
         project = project.filter(Project.removed_at.is_(None))
     project = project.first_or_404()
     workspace = project.workspace
+    if not workspace:
+        abort(404)
     if not is_active_workspace(workspace):
         abort(404, "Workspace doesn't exist")
     if not permission.check(project, current_user):
--- a/server/mergin/sync/permissions.py
+++ b/server/mergin/sync/permissions.py
@@ -157,15 +157,24 @@ class ProjectPermissions:
             return ProjectRole.READER
         return None
 
+def is_inactive_workspace(workspace):
+    """
+    Checks if the given workspace is active and if the current user is an admin or anonymous.
+
+    Args:
+        workspace (Workspace): The workspace to check.
+
+    Returns:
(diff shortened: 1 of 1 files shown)


server/mergin/sync/permissions.py
```diff
@@ -157,12 +157,24 @@ def get_user_project_role(
             return ProjectRole.READER
         return None
 
+def is_inactive_workspace(workspace):
```
harminius, line 160: I'd prefer a positive funciton name: `is_active_workspace`

server/mergin/sync/permissions.py
```diff
@@ -157,12 +157,24 @@ def get_user_project_role(
...
+        workspace (Workspace): The workspace to check.
+
+    Returns:
+        bool: True if the workspace is not active and the current user is not an admin or is anonymous, False otherwise.
+    """
+    is_admin = not current_user.is_anonymous and current_user.is_admin
     if not workspace:
         abort(404)
-    if not workspace.is_active and not current_user.is_admin:
+    return not workspace.is_active and (not is_admin or is_anonymous)
```
harminius, line 173: `is_anonymous` -> `current_user.is_anonymous`?

harminius:
It is worth adding some tests.

MarcelGeo:
> It is worth adding some tests.

Sure. It is hard to simulate AnonymouUser in tests ..

MarcelGeo:
Tests added in private repo :brain:

server/mergin/sync/permissions.py
```diff
@@ -158,11 +158,25 @@ def get_user_project_role(
...
-    workspace = current_app.ws_handler.get_by_name(ws)
+def is_active_workspace(workspace):
+    """
+    Checks if the given workspace is active and if the current user is an admin or anonymous.
+
+    Args:
+        workspace (Workspace): The workspace to check.
+
+    Returns:
+        bool: True if the workspace is not active and the current user is not an admin or is anonymous, False otherwise.
```
harminius, line 169: Update description, please

harminius (approved):
Looks good.

server/mergin/sync/permissions.py
```diff
@@ -158,11 +158,25 @@ def get_user_project_role(
...
+    """
     if not workspace:
         abort(404)
```
varmar05, line 172: this is not bool, also we do not need such check, I assume we won't pass None in arg

server/mergin/sync/permissions.py
```diff
@@ -158,11 +158,25 @@ def get_user_project_role(
         return None
 
 
-def require_project(ws, project_name, permission):
-    workspace = current_app.ws_handler.get_by_name(ws)
+def is_active_workspace(workspace):
```
varmar05, line 161: do we need this function at all? there is already similar check https://github.com/MerginMaps/server/blob/gh%232381-prevent-download-anonymous-inactive-ws/server/mergin/sync/workspace.py#L68

we would only need to handle current_user properly

MarcelGeo: I am not sure, this just abstracted code from require_project and required_project_by_uuid .

server/mergin/sync/permissions.py
MarcelGeo: https://github.com/MerginMaps/server/pull/239/commits/b267f71681394711a0d6c30b38794efdba8a374c

server/mergin/sync/permissions.py
MarcelGeo: https://github.com/MerginMaps/server/pull/239/commits/b267f71681394711a0d6c30b38794efdba8a374c

server/mergin/sync/permissions.py
varmar05: I am only worried we use various similar functions for doing the same thing and most probably not used everywhere - we might want to revisit this at some point.

varmar05 (approved):

server/mergin/sync/permissions.py
MarcelGeo: Thanks :) But currently I am no table to do some changes there .

PR #245, opened 2024-07-18
Add invitatation attribute with expires_at field to ProjectAccessDeta…

harminius:
Fixes https://github.com/MerginMaps/server-private/issues/2442

We forgot taht in EE/SaaS this CE code has to know about invitations which are not present in CE.

Problem:

In SaaS, Project Collaborators tab we ddn't know when the invitation is expiring.

Solution

Add `expires_at` field for `invitation` attribute of the project access schema.

image


MarcelGeo:
Hmm , we introduced project_access method to do this job ... in project_access method ... @varmar05 creating new issue , where this method could return modified project access in different env.

MarcelGeo (approved):

PR #263, opened 2024-08-12
Optimize handling of project files

varmar05:
Optimization for getting version files, particularly the latest one by:
- split file paths to separate table for easier lookup, optimize db query
- cache latest version files in new table on demand


MarcelGeo (approved):
After call discussion looks ok to me.

harminius:
@varmar05 did you run ee & service tests for this? 

I think there should be a PR to fix workspace tests (https://github.com/MerginMaps/server-private/blob/2024.5.2/server/test_ee/test_workspace.py#L64)

PR #264, opened 2024-08-13
Set correct Content-Type for responses with custom error

harminius:
Fixes [#2459](https://github.com/MerginMaps/server-private/issues/2459)

Error handling in py-client was done here: https://github.com/MerginMaps/python-api-client/pull/209

Problem:
The 'incorrect' `Content-Type` is when response is not created with `abort` command. This happens when we want some custom attributes in error response. 

Solution:
Add desirable `Content-Type` with after_request decorator to all custom error responses.

Alternative solution could be to set `problem+json` for `Content-Type` when returning the request in the method.

### DIFF
--- a/server/mergin/app.py
+++ b/server/mergin/app.py
@@ -396,6 +396,15 @@ def create_app(public_keys: List[str] = None) -> Flask:
 
         return response
 
+    @application.after_request
+    def set_custom_error_header(response):
+        """Clients (e.g. plugin) expect error `Content-Type` in response header.
+        Some responses with custom error lacks it."""
+        if response.status_code == 422:
+            response.headers["Content-Type"] = "application/problem+json"
+
+        return response
+
     # we need to register default handler to be accessible within app
     application.ws_handler = GlobalWorkspaceHandler()
     application.project_handler = ProjectHandler()
--- a/server/mergin/sync/public_api.yaml
+++ b/server/mergin/sync/public_api.yaml
@@ -272,7 +272,7 @@ paths:
         "422":
           description: Project access update could not be processed
           content:
-            application/json:
+            application/json+problem:
               schema:
                 anyOf:
                   - $ref: "#/components/schemas/UpdateProjectAccessError"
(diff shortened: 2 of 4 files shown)


server/mergin/app.py
```diff
@@ -396,6 +396,15 @@ def log_bad_request(response):
 
         return response
 
+    @application.after_request
+    def set_custom_error_header(response):
+        """Clients (e.g. plugin) expect error `Content-Type` in response header.
+        Some responses with custom error lacks it."""
+        if response.status_code == 422:
```
varmar05, line 403: is this only for our custom errors or all aborts, like 404, 409, ...

server/mergin/sync/public_api.yaml
```diff
@@ -272,7 +272,7 @@ paths:
         "422":
           description: Project access update could not be processed
           content:
-            application/json:
+            application/problem+json:
```
varmar05, line 275: let's grep all yaml files (even in EE/cloud)

server/mergin/app.py
harminius: when the response is created with aborter, it has correct type

however, custom error responses are not always created with abort() as we want some custom attributes to add there
e.g. https://github.com/MerginMaps/server/blob/master/server/mergin/sync/public_api_controller.py#L662

We also return some custom errors in `207` but I'm not sure we want to change Content-Type headers for that.

server/mergin/sync/public_api.yaml
harminius: New commit for the remaining CE `422` custom error response: https://github.com/MerginMaps/server/pull/264/commits/3906205eccb96d08f52d387803efeb9566b2ed12

Workspace yaml: https://github.com/MerginMaps/server-private/pull/2485
Service has no `422` error responses.

server/mergin/app.py
varmar05: ah ok, thanks

207 is success-like, I would not worry about that one

varmar05 (approved):

PR #268, opened 2024-08-15
Update backend packages

harminius:
Fixes: https://github.com/MerginMaps/server-private/issues/2330

Update sqlalchemy to `1.4.53`

Update flask to `2.2.5`

Resolve some dependabot issues and deprecated methods.

Update other unimportant packages.

`connexion 2.14 unfortunately requires `flask <2.3` so we cannot easily upgrade to newer versions.

image

Summary of changes:
I found no new relevant feature we could benefit from. Most updates are some bugfixes and usually Python 12 compatibility.


harminius:
Before this PR the number of warning in tests was: `54802`
In this PR: `40956`
:sweat_smile:

varmar05 (approved):

PR #269, opened 2024-08-15
Version files reverse search

varmar05:
Add possibility to calculate version files from the latest version. The new method should address performance issues for projects with large history and many files.

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -240,9 +240,9 @@ class Project(db.Model):
         db.session.execute(
             upload_table.delete().where(upload_table.c.project_id == self.id)
         )
-        self.access.owners = self.access.writers = self.access.editors = (
-            self.access.readers
-        ) = []
+        self.access.owners = (
+            self.access.writers
+        ) = self.access.editors = self.access.readers = []
         access_requests = (
             AccessRequest.query.filter_by(project_id=self.id)
             .filter(AccessRequest.status.is_(None))
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -803,8 +803,8 @@ class ProjectVersion(db.Model):
         """
         return "v" + str(name)
 
-    @property
-    def files(self) -> List[ProjectFile]:
+    def _files_forward_search(self):
+        """Calculate version files using lookup from the first version"""
         query = f"""
             WITH latest_changes AS (
                 SELECT
@@ -812,10 +812,11 @@ class ProjectVersion(db.Model):
                     max(pv.name) AS version
                 FROM
                     project_version pv
-                    LEFT OUTER JOIN file_history fh ON fh.version_id = pv.id
-                    LEFT OUTER JOIN project_file_path fp ON fp.id = fh.file_path_id
+                LEFT OUTER JOIN file_history fh ON fh.version_id = pv.id
+                LEFT OUTER JOIN project_file_path fp ON fp.id = fh.file_path_id
                 WHERE
-                    pv.project_id = :project_id AND pv.name <= :version
+                    pv.project_id = :project_id
+                    AND pv.name <= :version
                 GROUP BY
                     fp.id
             )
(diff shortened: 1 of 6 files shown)


MarcelGeo:
@mostlyAtNight :eagle:

server/mergin/sync/models.py
```diff
@@ -833,7 +834,71 @@ def files(self) -> List[ProjectFile]:
...
+            WHERE fh.change != 'delete'
+            ORDER BY fp.path;
+        """
+        params = {"project_id": self.project_id, "version": self.name}
+        return db.session.execute(query, params).fetchall()
+
+    @property
+    def files(self) -> List[ProjectFile]:
+        # return from cache
+        if self.name == self.project.latest_version:
```
MarcelGeo, line 895: Minor: IMO - do this logic in Project.files ... do not return any project staff from ProjectVersion ... :)

server/mergin/sync/models.py
```diff
@@ -833,7 +834,71 @@ def files(self) -> List[ProjectFile]:
...
+                    SELECT DISTINCT
+                        fh.file_path_id AS file_id
+                    FROM project_version pv
+                    LEFT OUTER JOIN file_history fh ON fh.version_id = pv.id
+                    WHERE
+                        pv.project_id = :project_id
+                        AND pv.name > :version
+                        AND fh.change = 'delete'
+                    -- union with current files
+                    UNION
```
MarcelGeo, line 854: Just question: Are you sure with ```UNION``` or you want to merge It with ```UNION ALL``` without any deduplication?


    UNION: only keeps unique records
    UNION ALL: keeps all records, including duplicates

server/mergin/sync/models.py
```diff
@@ -833,7 +834,71 @@ def files(self) -> List[ProjectFile]:
             WHERE fh.change != 'delete';
         """
         params = {"project_id": self.project_id, "version": self.name}
-        result = db.session.execute(query, params).fetchall()
+        return db.session.execute(query, params).fetchall()
+
+    def _files_backward_search(self):
```
MarcelGeo, line 839: I think this is not **search** ... this is reconstruction / calculation. See also _files_forward_search.

MarcelGeo (commented):
Looks good, just minor comments. One question:

- where is logic for ordering of files reconstruction? I see there just ORDER BY fp.path.

server/mergin/sync/models.py
varmar05: not sure I understand - if someone ask for latest version I want to redirect it to project.files rather then calculate

varmar05:
> Looks good, just minor comments. One question:
> 
> * where is logic for ordering of files reconstruction? I see there just ORDER BY fp.path.

there is none, we did not have one but we can introduce it - order by path seems natural to me

server/mergin/sync/models.py
```diff
@@ -833,7 +834,71 @@ def files(self) -> List[ProjectFile]:
             WHERE fh.change != 'delete';
         """
         params = {"project_id": self.project_id, "version": self.name}
-        result = db.session.execute(query, params).fetchall()
+        return db.session.execute(query, params).fetchall()
```
mostlyAtNight, line 837: Is this due to the fact that we now return early whereas before we did something else perhaps with result and returned it later?

server/mergin/sync/models.py
```diff
@@ -833,7 +834,71 @@ def files(self) -> List[ProjectFile]:
...
+            WITH files_changes_before_version AS (
+                WITH files_candidates AS (
+                    -- files removed later (but created anytime)
+                    SELECT DISTINCT
+                        fh.file_path_id AS file_id
+                    FROM project_version pv
+                    LEFT OUTER JOIN file_history fh ON fh.version_id = pv.id
+                    WHERE
+                        pv.project_id = :project_id
+                        AND pv.name > :version
```
mostlyAtNight, line 851: Should this be > or >= ?

server/mergin/sync/models.py
```diff
@@ -833,7 +834,71 @@ def files(self) -> List[ProjectFile]:
...
-        result = db.session.execute(query, params).fetchall()
+        return db.session.execute(query, params).fetchall()
+
+    def _files_backward_search(self):
+        """Calculate version files using lookup from the last version"""
+        query = f"""
+            WITH files_changes_before_version AS (
+                WITH files_candidates AS (
+                    -- files removed later (but created anytime)
+                    SELECT DISTINCT
```
mostlyAtNight, line 845: Did you want the DISTINCT to apply to the union of the two selects or just this first one? By the way UNION (rather than UNION ALL) already ensures that the union is not adding duplicates to the first query

mostlyAtNight (commented):
Hi @varmar05, the changes look good - I have a few questions / minor things we can discuss but I cannot see any problems. Unfortunately my understanding of the backwards search SQL is not 100% but still, I cannot see any problems with it.

Have we run the assert forawrd==backward test on some sizeable projects with realistic changes?

mostlyAtNight (approved):
Looks good, thanks for the disucussion and let's do some version tests on a clone of prod so we're super-confident there too 👍

PR #274, opened 2024-09-03
Optimize file history queries using cached version names

varmar05:
Simplify queries of file history so that JOIN with project version table is not needed. 

Tested on selected project with up to 50x speed up.


MarcelGeo (approved):
Looks good :) 

Waiting for merge

PR #278, opened 2024-09-05
Use dedicated working directory for Geodiff actions

varmar05:
This PR introduces dedicated working directory for geodiff actions. Necessary files are first copied from live directory and the result of the action is copied back where expected. All temporary files are cleaned afterwards. If copy fails, there is a fallback option to perform the action in current project directory.

The idea behind it is to have a possibility to use dedicated local volumes (with fast disks) for the manipulation with .gpkg files. There should be significant performance improvement comparing to live projects directory often being NFS.

Separation of working directory for files manipulation and storage for live files also opens a way to introduce different storage backend in the future (e.g. S3).

Note: There is no default clean up job for mounted volumes - this should be handled on deployment level.


server/mergin/sync/storages/disk.py
```diff
@@ -124,6 +132,26 @@ def _logger_callback(level, text_bytes):
...
+        """
+        file_path = file.replace(self.project_dir, "").lstrip(os.path.sep)
+        file_copy = os.path.join(self.geodiff_working_dir, file_path)
+        try:
+            copy_file(file, file_copy)
+            yield file_copy
+        except OSError as e:
+            logging.warning(f"Copy to geodiff dir failed: {str(e)}")
+            # fallback to live directory
+            file_copy = file + "_tmp"
```
MarcelGeo, line 149: Why we need this temp copy?

server/.test.env
```diff
@@ -19,3 +19,4 @@ INPUTAPP_API_KEY=not-secret-key
 GLOBAL_WORKSPACE='mergin'
 GLOBAL_STORAGE=104857600
 COLLECT_STATISTICS=0
+GEODIFF_WORKING_DIR=/tmp/geodiff
```
MarcelGeo, line 22: We need probably tune run_mergin to accept or not this variable, hm?

server/mergin/sync/storages/disk.py
```diff
@@ -112,6 +113,13 @@ def __init__(self, project):
         self.project_dir = self._project_dir()
         self.geodiff = GeoDiff()
         self.gediff_log = io.StringIO()
+        self.geodiff_working_dir = os.path.abspath(
```
MarcelGeo, line 116: Just wondering about CE/EE here. Do we want to use this functionality on the background in every instance? We can test If GEODIFF_WORKING_DIR exists and If not -> do not copy data somewhere to temp dir on the local images. It could be source of issues.

server/mergin/sync/storages/disk.py
```diff
@@ -214,56 +242,61 @@ def apply_diff(
...
-        db.session.add(gh)
-        return Ok(
-            (
-                checksum,
-                os.path.getsize(patchedfile),
-            )
-        )
+        with self.geodiff_copy(changeset) as changeset_tmp, self.geodiff_copy(
+            basefile
+        ) as patchedfile_tmp:
```
MarcelGeo, line 251: this is basefile_tmp no?

server/mergin/sync/storages/disk.py
```diff
@@ -278,22 +311,40 @@ def construct_diff(
...
+                    path=diff_name,
+                    checksum=generate_checksum(changeset_tmp),
+                    size=os.path.getsize(changeset_tmp),
+                    location=os.path.join(v_name, mergin_secure_filename(diff_name)),
+                )
+                copy_file(changeset_tmp, changeset)
+                return Ok(diff_file)
+            except (GeoDiffLibError, GeoDiffLibConflictError) as e:
+                # diff is not possible to create - file will be overwritten
+                move_to_tmp(changeset)
```
MarcelGeo, line 344: This is performed twice in finnally.

MarcelGeo (commented):
Just questions

server/mergin/sync/storages/disk.py
```diff
@@ -124,6 +132,26 @@ def _logger_callback(level, text_bytes):
...
+        """Copy project file from live storage to geodiff temp storage for further actions.
+        If file cannot be copied then as a fallback it creates _tmp file next to original file.
+        Temporary copy is removed on exit.
+        """
+        file_path = file.replace(self.project_dir, "").lstrip(os.path.sep)
+        file_copy = os.path.join(self.geodiff_working_dir, file_path)
+        try:
+            copy_file(file, file_copy)
+            yield file_copy
+        except OSError as e:
```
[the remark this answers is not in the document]
varmar05: since we are only copying the file I think this should be enough - mostly for the case of full disk error

server/mergin/sync/storages/disk.py
varmar05: this will be copy of _any_ original file passed to geodiff (yes, for some cases we might not need it - this is for the sake of consistency though)

server/.test.env
varmar05: True, this would be the case of any new introduced variable. On the other hand we have  sensible default so it is not a big deal.

server/mergin/sync/storages/disk.py
varmar05: It is implemented in the way it will work for all environments. Even on single machine host and local volume. I do not think it is a big deal to leave it there (no much overhead) and in fact I would favor working dir and storage dir separated in all cases.

server/mergin/sync/storages/disk.py
varmar05: no, it is not - we copy basefile as a new patched file because geodiff modifies this file with changeset

server/mergin/sync/storages/disk.py
varmar05: here we remove copy (if even created, L340) while in the finally clause we remove _tmp file

server/.test.env
MarcelGeo: This could lead to problems, If someone is using mountpoint for projects data and is having limits for local machines tmp?

server/mergin/sync/storages/disk.py
MarcelGeo: question here https://github.com/MerginMaps/server/pull/278#discussion_r1749895696

server/.test.env
varmar05: I think we should rather document it properly else we end up with if-else branches in code, do this or do that which I wanted to avoid.

These kind of deployments problems CE is vulnerable even now, is not it?

MarcelGeo: Yeah.. let's prepare PR to documentation :) It shoul be ok to maintain It there also :)

PR #279, opened 2024-09-06
Migration version author from str to user id

varmar05:
Change project_version.author -> author_id as FK in database. Due to historical data we keep new column as nullable.


server/mergin/tests/test_project_controller.py
```diff
@@ -2299,12 +2302,17 @@ def test_get_project_version(client, diff_project):
 
 
 def add_project_version(project, changes, version=None):
+    author = (
+        current_user
+        if current_user
+        else User.query.filter_by(username=DEFAULT_USER[0]).first()
```
MarcelGeo, line 2308: DEFAULT_USER s array?

MarcelGeo (changes requested):

server/migrations/community/1ab5b02ce532_version_author_name_to_user_id.py
```diff
@@ -0,0 +1,75 @@
+"""Migrage project version author name to user.id
+
+Revision ID: 1ab5b02ce532
+Revises: 57d0de13ce4a
```
MarcelGeo, line 4: Revises:


Revision ID: 1c23e3be03a3

server/mergin/tests/test_project_controller.py
MarcelGeo: It's tuple

MarcelGeo (approved):

PR #280, opened 2024-09-06
Keep project versions for removed projects

varmar05:
Now when project version table became lightweight we want to keep it even for removed projects. Files details are still removed.


harminius (approved):
Seems to be fine.

PR #281, opened 2024-09-09
Make sure file_history.diff is not null only when change is update_diff

varmar05:
- Make sure to pass correct diff to File history based on change type
- Introduce DB constraint

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -492,6 +492,17 @@ class FileHistory(db.Model):
             file_path_id,
             project_version_name.desc(),
         ),
+        db.CheckConstraint(
+            text(
+                """
+                CASE
+                    WHEN (change = 'update_diff') THEN diff IS NOT NULL
+                    ELSE diff IS NULL
+                END
+                """
+            ),
+            name="file_history_changes_with_diff",
+        ),
     )
 
     def __init__(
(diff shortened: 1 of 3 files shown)


server/mergin/sync/models.py
```diff
@@ -507,7 +518,7 @@ def __init__(
         self.size = size
         self.checksum = checksum
         self.location = location
-        self.diff = diff
+        self.diff = diff if diff is not None else null()
```
MarcelGeo, line 521: null() ? What's doing this?

MarcelGeo (approved):

server/migrations/community/1c23e3be03a3_add_file_history_diff_constraint.py
```diff
@@ -0,0 +1,40 @@
...
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    conn.execute(
+        sa.text(
+            """
+            ALTER TABLE file_history
+            ADD CONSTRAINT ck_file_history_changes_with_diff CHECK (
```
MarcelGeo, line 25: Again, name is different from name=file_history_changes_with_diff

MarcelGeo (changes requested):

MarcelGeo (approved):

PR #287, opened 2024-09-13
Redesign admin users

MarcelGeo:
**Redesign** :rose: 
- accounts 
image

- account detail with several advanced settings
image

- dialogs
image

Refactor :hammer: 
- dl -> dd element
- add font style for labels directly
- upgrade reset css for text-wrap
-


varmar05 (commented):
Looking good! I am just wondering if we want admin to change notifications / admin state in accounts details. I'd say former is done by user himself and the later should be rather done via server side command as it is kind of dangerous action.

PR #289, opened 2024-09-17
Redesign admin projects

MarcelGeo:
Redesign :rose: 
- Projects table
image
- Project detail in new light version without any copied functionality from dashboard - it's simple as possible.
image

- Dialogs
image

:exclamation: TODO:
- merge after #287


server/mergin/sync/workspace.py
```diff
@@ -248,16 +248,16 @@ def filter_projects(
     def workspace_count():
         return 1
 
-    def projects_query(self, name=None, workspace=None):
+    def projects_query(self, like: str = None):
```
varmar05, line 251: Is this used only in private API?

web-app/packages/admin-lib/src/modules/admin/views/ProjectView.vue
```diff
@@ -0,0 +1,222 @@
...
+                >
+              </template>
+            </div>
+            <div class="flex-shrink-0">
+              <PButton
+                @click="confirmPublicPrivate()"
+                severity="secondary"
+                data-cy="settings-public-btn"
+                :label="
+                  project?.access?.public ? 'Make private' : 'Make public'
```
varmar05, line 74: I would not let admin change this as I do not see a use case for it. Let's make it read-only.

web-app/packages/admin-lib/src/modules/admin/views/ProjectView.vue
```diff
@@ -0,0 +1,222 @@
...
+            <div class="flex-grow-1">
+              <p class="title-t3">Delete project</p>
+              <span class="paragraph-p6 opacity-80"
+                >Deleting this project will remove it and all its data. This
+                action cannot be undone.</span
+              >
+            </div>
+            <div class="flex-shrink-0">
+              <PButton
+                @click="confirmDelete"
```
varmar05, line 96: Do we have both, soft and hard delete options?

varmar05:
For project details could you @MarcelGeo liaise with @erpas about what is actually useful to have there (e.g. some aggregates like size, num of files, version, access, ...)?

server/mergin/sync/workspace.py
MarcelGeo: Yeah, I found just one usage of ws_handler.projects_query ... and that's in private controller for ```list_projects``` ...

web-app/packages/admin-lib/src/modules/admin/views/ProjectView.vue
MarcelGeo: I intrdoduced here hard delete only option. I am not sure If we want to have both in admin.

MarcelGeo:
> For project details could you @MarcelGeo liaise with @erpas about what is actually useful to have there (e.g. some aggregates like size, num of files, version, access, ...)?

Yeah ... In previous version too many options was not running. We have to make discussion with some users of what they want here.

PR #290, opened 2024-09-18
Change constraint name

varmar05:
Change constraint name to be `ck_file_history_changes_with_diff` and inline with alembic migration.


MarcelGeo (approved):
Nice

PR #294, opened 2024-09-20
Add workspace_id to paginated projects

MarcelGeo:
:exclamation: 

- merge after #293


server/mergin/sync/schemas.py
```diff
@@ -316,6 +317,12 @@ def _workspace_name(self, obj):
             name = obj.Project.workspace.name
         return name
 
+    def _workspace_id(self, obj):
```
varmar05, line 320: workspace_id is a column in project db, why this method?

varmar05 (changes requested):

varmar05 (approved):

PR #298, opened 2024-09-26
Add handling for os.renames on different filesystems to avoid OSError

varmar05:
In case GEODIFF _WORKING_DIR, LOCAL_PROJECTS_DIR and TEMP_DIR are not on the same filesystem we need to fallback to other renaming convention - use the same FS and only add a marker `delete-me`.

Normally this should not happen, but for certain deployments on k8s it may be beneficial to have various volumes of different types mounted. In that case a custom clean up job is necessary - this needs to be DOCUMENTED properly!


varmar05:
https://github.com/MerginMaps/docs/issues/505

MarcelGeo (approved):

PR #308, opened 2024-10-25
Enforce valid email address for User email

harminius:
resolves [#2537](https://github.com/MerginMaps/server-private/issues/2537)

Problem:
Users can register with an email containing 'invalid' characters, which causes (our) email service providers' API to crash. 

Solution:
Let's extend the existing email validator to disallow some additional characters.


server/mergin/auth/forms.py
```diff
@@ -64,6 +65,11 @@ class RegisterUserForm(FlaskForm):
     def validate(self):
         if not super().validate():
             return False
+        if not is_email_address_valid(self.email.data):
```
varmar05, line 68: Is this the only form, what about email update? Maybe we can create custom email validator?

server/mergin/auth/forms.py
```diff
@@ -64,6 +65,11 @@ class RegisterUserForm(FlaskForm):
...
+            self.email.errors.append(
+                "Email address contains non-ASCII characters. Only ASCII emails are allowed."
```
varmar05, line 70: I'd double check with Tomas for message.

server/migrations/community/1a6175c78a10_email_address_format_check.py
```diff
@@ -0,0 +1,28 @@
...
+Revises: 1ab5b02ce532
+Create Date: 2024-10-17 15:13:11.360991
+
+"""
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "1a6175c78a10"
```
varmar05, line 13: it can be bit misleading that it also starts with `1a`

varmar05 (changes requested):

server/mergin/auth/models.py
```diff
@@ -35,6 +35,10 @@ class User(db.Model):
     __table_args__ = (
         db.Index("ix_user_username", func.lower(username), unique=True),
         db.Index("ix_user_email", func.lower(email), unique=True),
+        CheckConstraint(
```
varmar05, line 38: This will fail on checking so-called anonymized email with pattern "deleted_\d{13}". Can we exclude it? 
Else we would need to update our deleting strategy and migrate db.

MarcelGeo:
I am not confident with this change and multi language support.

server/mergin/auth/forms.py
```diff
@@ -29,6 +30,11 @@ def username_validation(form, field):
         )
 
 
+def email_format(form, field):
```
varmar05, line 33: how about extending Email() validator?

server/mergin/auth/models.py
```diff
@@ -7,7 +7,7 @@
 from typing import List, Optional
 import bcrypt
 from flask import current_app, request
-from sqlalchemy import or_, func
+from sqlalchemy import or_, func, CheckConstraint
```
varmar05, line 10: do we need CheckConstraint?

server/mergin/auth/utils.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (C) Lutra Consulting Limited
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
+
+import re
+
+
+def contains_email_invalid_characters(email: str) -> bool:
+    """Check for invalid characters for the email address"""
+    invalid_characters = r"[ ;:’\–\—|,]"
```
varmar05, line 10: What about two `.` in username, how is this handled?

Do we have some sources about 'standard' emails validations?

server/mergin/auth/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    return bool(re.search(invalid_characters, email))
+
+
+def is_utf8(text: str) -> bool:
+    """Check if given text is in UTF-8 encoding"""
+    try:
+        text.encode("utf-8").decode("utf-8")
+        return "�" not in text
```
varmar05, line 18: Is this the best way?

varmar05 (changes requested):

varmar05:
@harminius I suggest also to check db against invalid characters and non-utf8 mess, if you can pass me the queries

server/mergin/auth/forms.py
```diff
@@ -42,6 +42,18 @@ def username_validation(form, field):
             raise ValidationError(error)
 
 
+class ExtendedEmail(Email):
+    """Custom email validation extending WTForms email validation"""
```
varmar05, line 46: can we mention here what chars are invalid just for sake of clarity?

server/mergin/tests/test_auth.py
```diff
@@ -889,3 +885,31 @@ def test_server_usage(client):
...
+    ("user1", 201),  # no problem
+    ("user\260", 400),  # disallowed character
+    ("usér", 201),  # non-ascii character
+    ("user\\", 400),  # disallowed character
+    ("日人日本人", 201),  # non-ascii character
+    ("user|", 400),  # vertical bar
+    ("us er", 400),  # whitespace
+    ("us,er", 400),  # comma
+    ("us—er", 400),  # dash
+    ("us'er", 400),  # apostrophe
```
varmar05, line 900: how about these use cases:
- whitespace around word
- starting with .
- using . in word

varmar05 (changes requested):

varmar05 (approved):

PR #310, opened 2024-10-29
Introduce gevent request timeout

varmar05:
Currently we do not have any bounds in case we run server with gevent worker type. In case client gave up on connection server still might be processing request. Also if request is greenlet safe (as it should be!) it can take long time to finish since it yields to gevent hub and standard gunicorn timeout is not applied here.

This PR introduces custom timeout via middleware to cut the request early enough in case of gevent. There are two new related variables `GEVENT_WORKER` and `GEVENT_REQUEST_TIMEOUT`.

Logs from a dummy test
```
[ERROR] [23194] Gevent worker: Request /ping timed out
[ACCESS] - GET /ping  HTTP/1.1 502 11 - "Mozilla/5.0 (X11; Linux x86_64) ..." 1003113 <23194>
```


server/application.py
```diff
@@ -9,14 +9,17 @@
 # Modules that had direct imports (NOT patched): ['urllib3.util, 'urllib3.util.ssl']"
 # which comes from using requests (its deps) lib in webhooks
 
-import os
-from random import randint
+from mergin.config import Configuration as MainConfig
 
-if not os.getenv("NO_MONKEY_PATCH", False):
```
MarcelGeo, line 15: Do we want to refactor this NO_MONKEY_PATCH variable in whole system?

MarcelGeo (changes requested):

server/application.py
```diff
@@ -9,14 +9,17 @@
...
+if MainConfig.GEVENT_WORKER:
     import gevent.monkey
+    import psycogreen.gevent
+
+    gevent.monkey.patch_all()
+    psycogreen.gevent.patch_psycopg()
 
-    gevent.monkey.patch_all(subprocess=True)
```
mostlyAtNight, line 18: You're happy with the change in syntax of patching here? (patching without subprocess=True and explicitly patching psycopy nowadays)

server/config.py
```diff
@@ -59,12 +47,7 @@
 
 max_requests_jitter = 5000
 
-
-def do_post_fork(server, worker):
-    patch_psycopg()
-
-
-post_fork = do_post_fork
```
mostlyAtNight, line 67: What's with the change in post_fork busines? Feel free to chat with me about this rather than responding here.

server/mergin/app.py
```diff
@@ -105,9 +107,24 @@ def update_obj(self, obj):
...
+        self.app = app
+
+    def __call__(self, environ, start_response):
+        request = Request(environ)
+        try:
+            with gevent.Timeout(Configuration.GEVENT_REQUEST_TIMEOUT):
+                return self.app(environ, start_response)
+        except gevent.Timeout:
+            logging.error(f"Gevent worker: Request {request.path} timed out")
+            resp = Response("Bad Gateway", mimetype="text/plain", status=502)
```
mostlyAtNight, line 123: Would a [504](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) not make more sense here?

server/mergin/tests/test_middleware.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pytest.mark.parametrize("use_middleware", [True, False])
+def test_use_middleware(use_middleware):
+    """Test using middleware"""
+    Configuration.GEVENT_WORKER = use_middleware
+    Configuration.GEVENT_REQUEST_TIMEOUT = 1
+    application = create_simple_app()
+
+    def ping():
```
mostlyAtNight, line 19: A general point about testing (no need to modify anything here) but how sure are we that gevent has actually terminated the underlying request when it times out? It would be useful to create a test endpoint that (when active) generates high CPU load, perhaps with a tiny sleep so it's not completely blocking, and test that when a request to it times out, that CPU load does indeed drop back down.

mostlyAtNight (commented):
a few comments to go through but looks great

server/application.py
mostlyAtNight: Spoken about, sounds all good

server/config.py
mostlyAtNight: It turns out we could not find any documentation suggesting why this was done 'post_fork'

server/application.py
varmar05: Yes, I will create PR for it.

server/mergin/app.py
varmar05: done

server/mergin/tests/test_middleware.py
varmar05: I did test with both, blocking and non-blocking functions generating high CPU load. In case of gevent timeout, load dropped after 504 return code. In case of blocking action, 200 was returned but whole process had to finish with high cpu load.

mostlyAtNight: Perfect!

MarcelGeo (approved):

PR #316, opened 2024-10-31
add mohtly contributors count to send_statistics

MarcelGeo:


server/mergin/sync/workspace.py
```diff
@@ -248,6 +248,10 @@ def filter_projects(
     def workspace_count():
         return 1
 
+    @staticmethod
+    def monthly_contributors_count():
+        return 0
```
varmar05, line 253: Should not be some count over project version (pushes)? Or we do not care here.

MarcelGeo: We do not have contributors table here. But we can calculate it from project versions

tomasMizera (approved):
I am not familiar with sqlalchemy syntax, but it looks good to me

PR #319, opened 2024-11-05
Roles refactoring: Add new project_member table, update getter/setter…

MarcelGeo:
Resolves https://github.com/MerginMaps/server-private/issues/2586

### DIFF
--- a/server/mergin/auth/schemas.py
+++ b/server/mergin/auth/schemas.py
@@ -33,17 +33,16 @@ class UserProfileSchema(ma.SQLAlchemyAutoSchema):
 
     def _has_project(self, obj):
         # DEPRECATED functionality - kept for the backward-compatibility
-        from ..sync.models import Project, ProjectAccess
+        from ..sync.models import ProjectUser, Project
 
         ws = current_app.ws_handler.get_by_name(obj.user.username)
         if ws:
             projects_count = (
-                Project.query.filter(Project.creator_id == obj.user.id)
+                Project.query.join(ProjectUser)
+                .filter(Project.creator_id == obj.user.id)
                 .filter(Project.removed_at.is_(None))
-                .filter_by(workspace_id=ws.id)
-                .filter(
-                    Project.access.has(ProjectAccess.readers.contains([obj.user.id]))
-                )
+                .filter(Project.workspace_id == ws.id)
+                .filter(ProjectUser.user_id == obj.user.id)
                 .count()
             )
             return projects_count > 0
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -67,7 +67,7 @@ class Project(db.Model):
     removed_by = db.Column(
         db.Integer, db.ForeignKey("user.id"), nullable=True, index=True
     )
-    public = db.Column(db.Boolean, default=False, index=True)
+    public = db.Column(db.Boolean, default=False, index=True, nullable=False)
 
     creator = db.relationship(
         "User", uselist=False, backref=db.backref("projects"), foreign_keys=[creator_id]
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -67,9 +67,14 @@ class Project(db.Model):
     removed_by = db.Column(
         db.Integer, db.ForeignKey("user.id"), nullable=True, index=True
     )
+    public = db.Column(db.Boolean, default=False, index=True)
 
     creator = db.relationship(
         "User", uselist=False, backref=db.backref("projects"), foreign_keys=[creator_id]
+    )
+
+    project_users = db.relationship(
+        "ProjectUser", cascade="all, delete-orphan", back_populates="project"
     )
 
     __table_args__ = (db.UniqueConstraint("name", "workspace_id"),)
@@ -82,6 +87,7 @@ class Project(db.Model):
         self.workspace_id = workspace.id
         self.creator = creator
         self.latest_version = 0
+        self.public = kwargs.get("public", False)
         latest_files = LatestProjectFiles(project=self)
         db.session.add(latest_files)
 
@@ -240,9 +246,7 @@ class Project(db.Model):
         db.session.execute(
             upload_table.delete().where(upload_table.c.project_id == self.id)
         )
-        self.access.owners = self.access.writers = self.access.editors = (
-            self.access.readers
-        ) = []
+        self.project_users.clear()
         access_requests = (
             AccessRequest.query.filter_by(project_id=self.id)
             .filter(AccessRequest.status.is_(None))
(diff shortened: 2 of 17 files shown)


MarcelGeo (commented):
migration is not running without intarray extension in postgres.
possible alternative would be use EXCEPT operator

```sql
SELECT DISTINCT project_id, user_id, role FROM (SELECT
                        project_id AS project_id,
                        unnest(owners) AS user_id,
                        'owner' AS role
                FROM project_access
UNION
SELECT
    project_id,
    unnest(writer_ids) AS user_id,
    'writer' AS role
FROM (
    SELECT
        project_id,
        array(SELECT UNNEST(writers) EXCEPT SELECT UNNEST(owners)) AS writer_ids
    FROM project_access
) AS subquery
UNION
SELECT
    project_id,
    unnest(writer_ids) AS user_id,
    'writer' AS role
FROM (
    SELECT
        project_id,
        array(SELECT UNNEST(editors) EXCEPT SELECT UNNEST(array_cat(writers, owners))) AS writer_ids
    FROM project_access
) AS subquery) unioned
;
```

Another problem shoukd be with results of query, where multiple non unique users ids is returned. 


do we want to use intarray extension?

server/migrations/community/d02961c7416c_add_project_member_table.py
```diff
@@ -0,0 +1,235 @@
...
+            ondelete="CASCADE",
+        ),
+        sa.ForeignKeyConstraint(
+            ["user_id"],
+            ["user.id"],
+            name=op.f("fk_project_member_user_id_user"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint(
+            "project_id", "user_id", name=op.f("pk_project_member")
```
mostlyAtNight, line 46: Is it therefore true that there'll only ever be one row in this table, regardless of access level, for a given user on a given project?

server/migrations/community/d02961c7416c_add_project_member_table.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def downgrade():
+    op.create_table(
+        "project_access",
+        sa.Column("public", sa.BOOLEAN(), autoincrement=False, nullable=True),
+        sa.Column(
+            "owners",
+            postgresql.ARRAY(sa.INTEGER()),
+            server_default=sa.text("'{}'::integer[]"),
```
mostlyAtNight, line 66: This is different to the schema we run right now in production (we don't use a default)

mostlyAtNight (commented):
Hi guys, I've reviewed the migration script thoroughly but have not looked at any other part of the MR.

Some other thoughts:
- we should ensure an upgrade/downgrade cycle results in exactly the same states for the tables affected 
- do we need to grant any permissions on new tables, or these will be taken care of automaticlaly with default schema permissions?

server/migrations/community/d02961c7416c_add_project_member_table.py
```diff
@@ -0,0 +1,235 @@
...
+            name=op.f("fk_project_member_user_id_user"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint(
+            "project_id", "user_id", name=op.f("pk_project_member")
+        ),
+    )
+
+    op.add_column("project", sa.Column("public", sa.Boolean(), nullable=True))
+    op.create_index(op.f("ix_project_public"), "project", ["public"], unique=False)
```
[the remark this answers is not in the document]
mostlyAtNight: Feel free to ignore this

server/migrations/community/d02961c7416c_add_project_member_table.py
mostlyAtNight: Discussed and this should not matter

server/migrations/community/d02961c7416c_add_project_member_table.py
```diff
@@ -0,0 +1,235 @@
...
+                FROM project_access
+                UNION
+                SELECT
+                        project_id AS project_id,
+                        unnest(readers - editors - writers - owners) AS user_id,
+                        'reader' AS role
+                FROM project_access
+            )
+            INSERT INTO project_member (project_id, user_id, role)
+            SELECT DISTINCT project_id, user_id, role::project_role FROM members;
```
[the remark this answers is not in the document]
mostlyAtNight: It seems we are happy with this and will confirm with our upgrade/downgrade consistency check

server/migrations/community/d02961c7416c_add_project_member_table.py
```diff
@@ -0,0 +1,235 @@
...
+                        project_member
+                    GROUP BY project_id, role
+                )
+                SELECT
+                    o.project_id,
+                    o.users_ids AS owners,
+                    o.users_ids || w.users_ids AS writers,
+                    o.users_ids || w.users_ids || e.users_ids AS editors,
+                    o.users_ids || w.users_ids || e.users_ids || r.users_ids AS readers
+                FROM (SELECT * FROM agg WHERE role = 'owner') AS o
```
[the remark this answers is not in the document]
mostlyAtNight: Let's tackle this before doing the upgrade/downgrade check

server/mergin/auth/schemas.py
```diff
@@ -33,17 +33,16 @@ def get_disk_usage(self, obj):
...
         # DEPRECATED functionality - kept for the backward-compatibility
-        from ..sync.models import Project, ProjectAccess
+        from ..sync.models import ProjectUser, Project
 
         ws = current_app.ws_handler.get_by_name(obj.user.username)
         if ws:
             projects_count = (
-                Project.query.filter(Project.creator_id == obj.user.id)
+                Project.query.join(ProjectUser)
+                .filter(Project.creator_id == obj.user.id)
```
MarcelGeo, line 42: creator_id looks like new condition here. Do we need this check?

server/mergin/sync/models.py
```diff
@@ -253,24 +257,43 @@ def delete(self, removed_by: int = None):
...
+    EDITOR = "editor"
+    WRITER = "writer"
+    OWNER = "owner"
 
-        https://docs.python.org/3/library/enum.html#enum.EnumType.__members__
-        """
+    def __ge__(self, other):
+        """Compare project roles"""
         members = list(ProjectRole.__members__)
-        if members.index(self.name) < members.index(other.name):
```
MarcelGeo, line 270: Are we sure with this change of order?

MarcelGeo (commented):
Downgrade, upgrade working for my local data. I also checked if everything is ok with data.

server/migrations/community/d02961c7416c_add_project_member_table.py
```diff
@@ -0,0 +1,235 @@
...
+        ),
+    )
+
+    op.add_column("project", sa.Column("public", sa.Boolean(), nullable=True))
+    op.create_index(op.f("ix_project_public"), "project", ["public"], unique=False)
+
+    data_upgrade()
+
+    op.drop_table("project_access")
+    conn.execute(sa.text("DROP EXTENSION IF EXISTS intarray;"))
```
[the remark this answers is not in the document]
MarcelGeo: Then just create extension without dropping. I thinks there is not issue to have it in other deployments

server/mergin/auth/schemas.py
varmar05: creator_id was there before in filter, but in fact it is some legacy stuff for backward compatibility

server/migrations/community/d02961c7416c_add_project_member_table.py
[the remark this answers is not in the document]
varmar05: done

server/migrations/community/d02961c7416c_add_project_member_table.py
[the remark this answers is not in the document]
varmar05: adjusted to count all projects - as it was causing inconsistencies

mostlyAtNight (approved):
Looks all good from my side

server/mergin/auth/schemas.py
MarcelGeo: ok ... this looks like legacy method :)

PR #320, opened 2024-11-07
Create project in tests

harminius:
Minifix to send ProjectVersion.created signal when creating project in tests.

On this signal, a Workspace Contributor entry is created. It's useful in tests.

### DIFF
--- a/server/mergin/tests/utils.py
+++ b/server/mergin/tests/utils.py
@@ -10,7 +10,6 @@ import math
 from dataclasses import asdict
 from datetime import datetime
 
-import pysqlite3
 from flask import url_for, current_app
 import os
 from dateutil.tz import tzlocal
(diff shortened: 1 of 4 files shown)


server/mergin/tests/utils.py
```diff
@@ -88,6 +88,7 @@ def create_project(name, workspace, user, **kwargs):
     pv = ProjectVersion(p, 0, user.id, changes, "127.0.0.1")
     db.session.add(pv)
     db.session.commit()
+    project_version_created.send(pv)
```
varmar05, line 91: Should not we trigger signal from ProjectVersion constructor instead?

harminius: That would be probably the safest.

MarcelGeo: Yes

varmar05: Indeed, I suggest some refactoring here, best to set it in one place and remove elsewhere.

MarcelGeo (approved):

PR #321, opened 2024-11-07
Error in push finish

harminius:
Task: https://github.com/MerginMaps/server-private/issues/2558

Situation:
When migrating project files project without a file were given Null value to the file history id. We expect this field not be Null. The caching method which should update the field fails to do it because it cannot find a file associated with the project id.

Solution
When no files are found when caching the latest files, put an empty array in the column and don't keep it Null.

Bonus: 
Fix flask cmd to create a project.


server/mergin/sync/models.py
```diff
@@ -743,7 +743,9 @@ def __init__(
         latest_files_map = {
             fh.path: fh.id
             for fh in FileHistory.query.filter(
-                FileHistory.id.in_(self.project.latest_project_files.file_history_ids)
+                FileHistory.id.in_(
+                    self.project.latest_project_files.file_history_ids or {}
```
varmar05, line 747: Could you add a test?

varmar05 (changes requested):

server/mergin/sync/commands.py
```diff
@@ -28,6 +29,7 @@ def project():
     def create(name, namespace, user):  # pylint: disable=W0612
         """Create blank project"""
         workspace = current_app.ws_handler.get_by_name(namespace)
+        user = User.query.get(int(user))
```
varmar05, line 32: I would probably go with username

varmar05:
"Make file_history_ids not nullable?" It was meant to be nullable and updated only on demand (first pull / get info) as whole migration would have been very heavy. So probably we maybe want to cache files ids if it is null as a fix.

server/mergin/sync/commands.py
harminius: Done

server/mergin/sync/models.py
harminius: Done

harminius:
> "Make file_history_ids not nullable?" It was meant to be nullable and updated only on demand (first pull / get info) as whole migration would have been very heavy. So probably we maybe want to cache files ids if it is null as a fix.

Actually, I found out the latest project files are cached in this scenario.

That's why the problem is only in empty projects. The caching method finds no files to cache and keeps the column `null`.

varmar05:
> Actually, I found out the latest project files are cached in this scenario.
> 
> That's why the problem is only in empty projects. The caching method finds no files to cache and keeps the column `null`.

If caching method finds no files then I guess it should set it to empty array rather that NULL which should be only used if that row has never been cached.

PR #323, opened 2024-11-11
Gh#2610 enable assignment variable

MarcelGeo:
Introduce ENABLE_SUPERADMIN_ASSIGNMENT variable which enables / disables admin access button in admin account detail. 

:exclamation: 
- merge after #322


server/mergin/auth/controller.py
```diff
@@ -408,11 +408,18 @@ def update_user(username):  # pylint: disable=W0613,W0612
...
+        "ENABLE_SUPERADMIN_ASSIGNMENT"
+    ):
+        abort(400, "Unable to assign super admin role")
 
     user = User.query.filter_by(username=username).first_or_404("User not found")
     form.update_obj(user)
+
     # remove inactive since flag for ban or re-activation
     user.inactive_since = None
+    print("heeeere")
```
varmar05, line 421: rm

server/mergin/config.py
```diff
@@ -102,3 +102,7 @@ class Configuration(object):
     USER_SELF_REGISTRATION = config("USER_SELF_REGISTRATION", default=False, cast=bool)
     # build hash number
     BUILD_HASH = config("BUILD_HASH", default="")
+    # Allow changing access to admin panel
+    ENABLE_SUPERADMIN_ASSIGNMENT = config(
+        "ENABLE_SUPERADMIN_ASSIGNMENT", default=True, cast=bool
```
varmar05, line 107: I wonder if False is not a safer default.

varmar05 (changes requested):

server/mergin/config.py
MarcelGeo: For other deployemnts, It could be True by default. we can adjust it in our deployments,

varmar05 (approved):

PR #326, opened 2024-11-11
Flush to DB right after project is created

harminius:
Let's fix the tests. 

We need project object attributes and relations to be accessible right after it is created because when ProjectVersion is created is sends signal which expects it.

### DIFF
--- a/server/mergin/sync/public_api_controller.py
+++ b/server/mergin/sync/public_api_controller.py
@@ -193,6 +193,7 @@ def add_project(namespace):  # noqa: E501
 
         p = Project(**request.json, creator=current_user, workspace=workspace)
         p.updated = datetime.utcnow()
+        db.session.add(p)
         db.session.flush()
         pa = ProjectAccess(p, public=request.json.get("public", False))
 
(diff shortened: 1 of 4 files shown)


server/mergin/sync/public_api_controller.py
```diff
@@ -193,6 +193,7 @@ def add_project(namespace):  # noqa: E501
 
         p = Project(**request.json, creator=current_user, workspace=workspace)
         p.updated = datetime.utcnow()
+        db.session.flush()
```
varmar05, line 196: Is not enough just to call session.add here which would autoflush changes?

Also is this the only place which causes issues for project version created? How about clone/push/push finish?

varmar05 (changes requested):

server/mergin/sync/public_api_controller.py
harminius: Some tests are failing without flushing the DB.
However, it's probably a good practice to add the instance to the DB before flushing the DB.

The clone controller doesn't fail :man_shrugging: but let's add the project to DB right after initialization.

We do not create a new project when pushing changes.

varmar05 (approved):

PR #328, opened 2024-11-12
Revert "Create project in tests"

MarcelGeo:
Reverts MerginMaps/server#320


varmar05 (approved):
Better safe than sorry :-)

PR #332, opened 2024-11-18
Fix current access api

varmar05:


server/mergin/sync/public_api_controller.py
```diff
@@ -535,19 +556,19 @@ def get_projects_by_uuids(uuids):  # noqa: E501
...
-        # FIXME
-        # user_ids.extend(p.access.owners + p.access.writers + p.access.readers)
-        ws_ids.append(p.workspace_id)
+    ws_ids = [p.workspace_id for p in projects]
+    projects_ids = [p.id for p in projects]
     users_map = {
-        u.id: u.username for u in User.query.filter(User.id.in_(set(user_ids))).all()
+        u.id: u.username
+        for u in User.query.select_from(ProjectUser)
+        .join(User)
```
MarcelGeo, line 569: this join is necessary?

server/mergin/sync/public_api_controller.py
```diff
@@ -535,19 +556,19 @@ def get_projects_by_uuids(uuids):  # noqa: E501
...
     projects = (
         projects_query(ProjectPermissions.Read, as_admin=False)
         .filter(Project.id.in_(proj_ids))
         .all()
     )
-    for p in projects:
-        # FIXME
-        # user_ids.extend(p.access.owners + p.access.writers + p.access.readers)
-        ws_ids.append(p.workspace_id)
+    ws_ids = [p.workspace_id for p in projects]
```
MarcelGeo, line 564: Do we want to have unique workspace ids?

server/mergin/sync/public_api_controller.py
```diff
@@ -622,17 +643,23 @@ def get_paginated_projects(
         only_public,
     )
     result = projects.paginate(page, per_page).items
-    total = projects.paginate(page, per_page).total
+    # for count (done in subquery) we only need ids and then to do distinct for deduplications due to joins
+    # https://docs.sqlalchemy.org/en/20/faq/sessions.html#my-query-does-not-return-the-same-number-of-objects-as-query-count-tells-me-why
+    total = (
```
MarcelGeo, line 648: Why is this different total then paginate total?

server/mergin/sync/public_api_controller.py
```diff
@@ -622,17 +643,23 @@ def get_paginated_projects(
...
-        # FIXME
-        # user_ids.extend(p.access.owners + p.access.writers + p.access.readers)
-        pass
-
+    projects_ids = [p.id for p in result]
     users_map = {
-        u.id: u.username for u in User.query.filter(User.id.in_(set(user_ids))).all()
+        u.id: u.username
+        for u in User.query.select_from(ProjectUser)
+        .join(User)
```
MarcelGeo, line 660: similar to :up:

server/mergin/tests/test_project_controller.py
```diff
@@ -660,20 +654,15 @@ def test_update_project(client):
...
 
     # add tests user as writer to project
-    current_writers = [
-        u.id for u in project.project_users if u.role == ProjectRole.WRITER.value
-    ]
-    writers = [
-        u.username for u in User.query.filter(User.id.in_(current_writers)).all()
-    ]
-    data = {"access": {"writersnames": writers + [test_user.username]}}
+    data = {"access": {"writersnames": [test_user.username]}}
```
MarcelGeo, line 657: I think in previous version was this access accepting whole array of names. If you defined one username, one username was saved to it, or?

MarcelGeo (commented):
looks good just few questions.

server/mergin/sync/public_api_controller.py
varmar05: yes, it is, we need to get username 

no worries here, it is again just sqlalchemy api thing to return correct object

server/mergin/sync/public_api_controller.py
varmar05: yea, let's do it, not that it would affect much

server/mergin/sync/public_api_controller.py
varmar05: Unfortunately it is how sqlalchemy implements count() api, it does count with some subselect which in case of joins have some duplicates - I had to add join() to query to sort out proper access filters. The underlying change is that before we had 1:1 project-project_access now we have 1:n with project_member.
At the end, count() >= len(all()).

From all I read about this issue, the best way to overcome is to count in deduplicated result (therefore distinct()).

server/mergin/tests/test_project_controller.py
varmar05: it still does accept whole array, but I had to modify the test because now if we have for example owner, there is no writer (and thus no writernames). Previously if you set one writer - you would end up with one writer (which is still the case now). 

The logic is following, whatever you specify in the list, will be set - either by upgrade/downgrade or adding new member. If you omit some key, that role will not by handled (only downgrade possible). If there is single id in multiple keys, the highest role wins.

server/mergin/sync/public_api_controller.py
varmar05: cool, I rewritten the query to use subquery in favour of join and we do not need to worry about that anymore

PR #337, opened 2024-11-29
Roles refactor: New public API with reworked project roles

varmar05:
Added new public `/v2` endpoints for project members CRUD and new endpoint for registration of users.
Also small refactoring was done (e.g. moving workspace role definition from EE).


MarcelGeo (commented):
Looks good. Just minor suggestions and one question:

- do we want to use ProjectPermissions.Upload and other methods from project permissions class? I think we could rewrite it or remove.

server/mergin/app.py
```diff
@@ -177,7 +177,6 @@ def create_app(public_keys: List[str] = None) -> Flask:
         arguments={"title": "Mergin"},
         options={"swagger_ui": Configuration.SWAGGER_UI},
         validate_responses=True,
-        pythonic_params=True,
```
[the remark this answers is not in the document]
varmar05: I found it used only it this new v2 blueprint, so in functions params you pass `id_` instead of `id` but found it rather confusing. If we want it this way, then we should propagate it everywhere.

varmar05:
> 
> * do we want to use ProjectPermissions.Upload and other methods from project permissions class? I think we could rewrite it or remove.

I did not touch this one, as it is widely used. If we want further refactoring then I'd probably vote for removing it completely.

So far it is hanging there if we one day decide to introduce permissions besides roles.

server/mergin/sync/public_api_controller.py
```diff
@@ -216,6 +216,7 @@ def add_project(namespace):  # noqa: E501
         request.json["storage_params"] = {
             "type": "local",
             "location": generate_location(),
+            "public": request.json.get("public", False),
```
varmar05, line 219: @harminius I am bit confused here, what's the catch with this 'public' flag?

harminius: I'm just looking at it now too. This has no sense. 
Marcel wanted something like this `request.json["public"] = request.json.get("public", False)`
Which I think is not needed as `public` is set to False in init when not passed.
Let me remove the line.

harminius (approved):
Just curious why we split that project access PUT.

PR #343, opened 2024-12-13
Add editor limit hit blocking

varmar05:
This is CE part for integration of new editor limit hit blocking mechanism. PR adds:

- a new error type
- a revert mechanism for updates that were rejected

### DIFF
--- a/server/mergin/sync/public_api_controller.py
+++ b/server/mergin/sync/public_api_controller.py
@@ -680,10 +680,33 @@ def update_project(namespace, project_name):  # noqa: E501  # pylint: disable=W0
     """
     project = require_project(namespace, project_name, ProjectPermissions.Update)
     parsed_access = parse_project_access_update_request(request.json.get("access", {}))
+
+    # get current status for easier rollback
+    modified_user_ids = []
+    for role in list(ProjectRole.__reversed__()):
+        modified_user_ids.extend(parsed_access.get(role, []))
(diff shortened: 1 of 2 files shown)


server/mergin/sync/public_api_controller.py
```diff
@@ -680,11 +680,34 @@ def update_project(namespace, project_name):  # noqa: E501  # pylint: disable=W0
...
+        rejected_users = (
+            db.session.query(User.id)
+            .filter(User.email.in_(error.rejected_emails))
+            .all()
+        )
+        for user in rejected_users:
+            if current_permissions_map[user.id] is None:
+                project.unset_role(user.id)
+            else:
+                project.set_role(user.id, current_permissions_map[user.id])
```
MarcelGeo, line 708: Do we need also set role in case of rejecting?

MarcelGeo (changes requested):

server/mergin/sync/public_api_controller.py
varmar05: yes, we do need to revert to original permissions - this is the case user was e.g. reader and we rejected to promote to writer

varmar05: yes, we do need to revert to original permissions - this is the case user was e.g. reader and we rejected to promote to writer

MarcelGeo (approved):

PR #348, opened 2024-12-18
Use v2 api to get project collaborators and their project permission

harminius:
Use new API to list project collaborators. Add a function to calculate real project permission from the project and workspace role.

Screenshot from 2024-12-18 11-56-41


web-app/packages/lib/src/common/permission_utils.ts
```diff
@@ -209,3 +217,22 @@ export function getProjectPermissionByRoleName(
   }
   return mapper[roleName]
 }
+
+export function calculateProjectPermission(
+  project_role: ProjectRoleName,
+  workspace_role: WorkspaceRoleName
+): ProjectRoleName {
+  const mappedWorkspaceRole: ProjectRoleName =
```
MarcelGeo, line 225: I think we can use mapper here WorkspaceRole -> ProjectRole

web-app/packages/lib/src/common/permission_utils.ts
```diff
@@ -209,3 +217,22 @@ export function getProjectPermissionByRoleName(
...
+    workspace_role === 'admin' ? 'owner' : (workspace_role as ProjectRoleName)
+
+  if (project_role === 'none' && workspace_role === 'guest') {
+    return 'none'
+  }
+
+  const projectRoleIndex = ROLE_HIERARCHY.indexOf(project_role)
+  const workspaceRoleIndex = ROLE_HIERARCHY.indexOf(mappedWorkspaceRole)
+
+  return projectRoleIndex > workspaceRoleIndex
```
MarcelGeo, line 235: We need calculate project role only if it's null. If not null -> stay same as before. I think we do not need such logic.

web-app/packages/lib/src/modules/project/components/ProjectMembersTable.vue
```diff
@@ -152,27 +153,31 @@ const searchedItems = computed(() =>
...
-function roleUpdate(item: ProjectAccessDetail, value: ProjectRoleName) {
+function roleUpdate(item: ProjectCollaborators, value: ProjectRoleName) {
   projectStore.updateProjectAccess({
     projectId: projectStore.project.id,
     userId: item.id,
     data: { role: value }
   })
 }
 
+function ProjectPermission(item: ProjectCollaborators) {
```
MarcelGeo, line 172: Rename to "projectRole"

web-app/packages/lib/src/modules/project/types.ts
```diff
@@ -349,3 +350,11 @@ export interface ProjectAccessDetail {
   project_permission: ProjectRoleName
   name: string
 }
+
+export interface ProjectCollaborators {
```
MarcelGeo, line 354: ProjectCollaborator**s** -> ProjectCollaborator

MarcelGeo (changes requested):

web-app/packages/lib/src/common/permission_utils.ts
harminius: A workspace role higher than a project role is a valid scenario, no? See user id 6...

Screenshot from 2024-12-18 11-56-41

MarcelGeo: We have deal with @varmar05 , that BE will return new role attribute which will be calculated role. We can then use it in dropdown and project_role will be used just for add / update project collaborator handling.

server/mergin/sync/private_api_controller.py
```diff
@@ -305,26 +304,15 @@ def unsubscribe_project(id):  # pylint: disable=W0612
...
 @auth_required
 def update_project_access(id: str):
-    """Modify shared project access
+    """Modify the project's public flag
 
     :param id: Project uuid
     """
     project = require_project_by_uuid(id, ProjectPermissions.Update)
 
     if "public" in request.json:
```
varmar05, line 313: I think if it is the only parameter, let's make it required.

varmar05: Also we can change response type, and even change URI?

server/mergin/sync/schemas.py
```diff
@@ -361,7 +361,8 @@ class ProjectAccessDetailSchema(Schema):
     role = fields.String()
     username = fields.String()
     name = fields.String()
-    project_permission = fields.String()
+    workspace_role = fields.String()
+    project_role = fields.String()
```
varmar05, line 365: let's use Enums everytime we return role

varmar05 (changes requested):

server/mergin/sync/private_api_controller.py
harminius: Any hints on what to return and the new URI?

MarcelGeo: /app/projects/{id}/public { public: boolean } ??

harminius: -> none?

varmar05 (approved):

MarcelGeo (approved):

PR #356, opened 2025-01-15
Fix: public switch

MarcelGeo:


varmar05 (approved):

alexd0793:
As i don't have a labrl with test ok i will write here. Test OK 👯‍♂️

PR #358, opened 2025-01-16
Fix: Deactivate account in admin

MarcelGeo:
There was bad interface in FE, where is_admin and is_active were required.


varmar05 (approved):

alexd0793:
As i don't have a labrl with test ok i will write here. Test OK 👯‍♂️

PR #359, opened 2025-01-20
Fix shared context issues for security salts

varmar05:


.prod.env
```diff
@@ -62,6 +62,8 @@ SECRET_KEY=fixme
 #BEARER_TOKEN_EXPIRATION=3600 * 12  # in seconds
 
 #SECURITY_PASSWORD_SALT=NODEFAULT
+SECURITY_BEARER_SALT=fixme
+SECURITY_EMAIL_SALT=fixme
```
MarcelGeo, line 66: Move it under SECURITY_PASSWORD_SALT . 

Add some text to comment over variable.

server/mergin/auth/bearer.py
```diff
@@ -7,8 +7,7 @@
 from flask.sessions import TaggedJSONSerializer
 
 
-def decode_token(secret_key, token, max_age=None):
-    salt = "bearer-session"
```
MarcelGeo, line 11: ouch. Good spot.

server/mergin/auth/controller.py
```diff
@@ -315,7 +319,9 @@ def confirm_new_password(token):  # pylint: disable=W0613,W0612
 
 
 def confirm_email(token):  # pylint: disable=W0613,W0612
-    email = confirm_token(token)
+    email = confirm_token(
+        token, expiration=12 * 3600, salt=current_app.config["SECURITY_EMAIL_SALT"]
```
MarcelGeo, line 323: Do we want to change expiration here?

MarcelGeo (changes requested):

server/.test.env
```diff
@@ -20,3 +20,6 @@ GLOBAL_WORKSPACE='mergin'
 GLOBAL_STORAGE=104857600
 COLLECT_STATISTICS=0
 GEODIFF_WORKING_DIR=/tmp/geodiff
+SECURITY_BEARER_SALT='bearer'
+SECURITY_EMAIL_SALT='email'
+SECURITY_PASSWORD_SALT='password'
```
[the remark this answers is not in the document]
varmar05: If I did not I would have to set it inside of tests, I think it is better this way as it more similar to prod setup

server/mergin/auth/controller.py
varmar05: I am open to any suggestion. I deliberately set default (and password reset) short but here for email confirmation it can be more relaxed.

.prod.env
varmar05: why under, I keep alphabetical order for this group (not that it matters)

I would keep notes in docs rather than here.

server/mergin/auth/bearer.py
varmar05: yea, best to get rid of it once for all

.prod.env
MarcelGeo: Because ```#SECURITY_PASSWORD_SALT=NODEFAULT``` above is related to another variable :)

server/mergin/auth/controller.py
MarcelGeo: Could you add it to any constant like variable on the top?

server/mergin/.env
```diff
@@ -1,6 +1,8 @@
 GEODIFF_LOGGER_LEVEL="2"
 # only for dev - should be overwritten in production
 SECRET_KEY='top-secret'
+SECURITY_BEARER_SALT='top-secret'
```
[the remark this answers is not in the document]
MarcelGeo: https://github.com/MerginMaps/docs/pull/542

MarcelGeo (approved):

PR #363, opened 2025-01-23
Gh#2726 callhome stats export

MarcelGeo:
Resolves https://github.com/MerginMaps/server-private/issues/2726

BE Impl :hammer: 
- update send statistics celery to store data to new introduced table ```MerginStatistics```
- New endpoint for getting csv from table between dates  in format ```2024-01-01```

FE Impl :hammer: 
- new dialog to settings template with ability to export data

image

image


server/mergin/stats/tasks.py
```diff
@@ -72,6 +72,10 @@ def send_statistics():
         "editors": current_app.ws_handler.server_editors_count(),
     }
 
+    stats = MerginStatistics(data=data)
```
varmar05, line 75: for data format see comment about dataclass

server/mergin/stats/tasks.py
```diff
@@ -72,6 +72,10 @@ def send_statistics():
...
+    db.session.add(stats)
```
varmar05, line 76: I  think this is too late as it depends on mergin_info.last_reported  which is only updated on successful post to server. I suggest to move this to top and create its own condition when we want to commit stats.

server/mergin/tests/test_statistics.py
```diff
@@ -152,3 +163,56 @@ def test_server_updates(client):
         mock.side_effect = requests.exceptions.RequestException("Some failure")
         resp = client.get(url)
         assert resp.status_code == 400
+
+
+def test_download_report(app, client):
+    """Test download report endpoint"""
+    url = "/app/admin/report"
+    resp = client.get(url)
+    resp.status_code == 400
```
varmar05, line 172: hm, maybe we can have default, e.g. last month?

server/mergin/tests/test_statistics.py
```diff
@@ -152,3 +163,56 @@ def test_server_updates(client):
...
+
+    # bad date format
+    resp = client.get(f"{url}?date_from=2021-01-01T00:00:00&date_to=2021-01-01")
+    assert resp.status_code == 400
+
+    app.config["COLLECT_STATISTICS"] = True
```
varmar05, line 178: I wonder what should be affected by variable - collection of stats or only report to our servers?

server/migrations/community/37cf462d76ed_add_mergin_statistics_table.py
```diff
@@ -0,0 +1,44 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "37cf462d76ed"
+down_revision = "d02961c7416c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
varmar05, line 21: let's remove comments

varmar05 (changes requested):

web-app/packages/lib/src/common/date_utils.ts
```diff
@@ -102,3 +102,13 @@ export function formatRemainingTime(t2, t1 = new Date()) {
       return remainingFormatDuration(days, 'days')
   }
 }
+
+/**
+ * Calculates date from today to given months number in past.
+ * Doest not convert to UTC.
+ */
+export function getDateFromMonthsAgo(months) {
```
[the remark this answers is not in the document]
MarcelGeo: do not need lib for this logic ...

server/mergin/tests/test_statistics.py
MarcelGeo: not sure ...

server/mergin/stats/controller.py
```diff
@@ -3,10 +3,27 @@
...
+
+from mergin.auth.app import auth_required
+from mergin.stats.models import MerginStatistics
 
 from .config import Configuration
-from ..app import parse_version_string
+from ..app import parse_version_string, db
+
+
+class CsvTextBuilder(object):
```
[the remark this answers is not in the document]
MarcelGeo: if you have this class, you can simulate creating stream of csv rows from csv library.

server/mergin/stats/controller.py
```diff
@@ -29,3 +46,45 @@ def get_latest_version():
...
+        abort(400, "Invalid date format")
+
+    stats = (
+        db.session.query(MerginStatistics.data)
+        .filter(MerginStatistics.created_at.between(parsed_from, parsed_to))
+        .order_by(MerginStatistics.created_at.desc())
+        .all()
+    )
+    data = [stat.data for stat in stats]
+    # get columns for data, this is usefull when we will update data json format (removing columns, adding new ones)
```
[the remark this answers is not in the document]
MarcelGeo: yes, I am doing on this method. There could be issue with backward compatibility of data class and fields in data json.

web-app/packages/admin-lib/src/modules/admin/store.ts
```diff
@@ -325,6 +328,43 @@ export const useAdminStore = defineStore('adminModule', {
...
+    },
+
+    async downloadReport(payload: { from: Date; to: Date }) {
+      const notificationStore = useNotificationStore()
+
+      try {
+        const { from, to } = payload
+        const url = AdminApi.contructDownloadStatisticsUrl()
+        const date = new Date()
+        // we need to sanitize hours from custom range in Calendar picker, because it's 00:00:00 by default. If you convert it to UTC -> it could be previous day, which is not what you selected.
```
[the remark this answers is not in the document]
MarcelGeo: yes, because if you select custom range just one day (e.g. 22.1.2025) ... data picker is returning from 22.1.2025 00:00:00 and to 22.1.2025 and you are not able to select stats for current day. 2. issue is discussed - yeah, we do not need it, but it's more precise.

server/mergin/stats/controller.py
```diff
@@ -29,3 +46,45 @@ def get_latest_version():
...
+    builder = CsvTextBuilder()
+    writer = DictWriter(
+        builder,
+        fieldnames=sorted_columns,
+    )
+    writer.writeheader()
+    writer.writerows(data)
+    csv_data = "".join(builder.data)
+    response = make_response(csv_data)
+    response.headers["Content-Disposition"] = f"attachment; filename=usage-report.csv"
```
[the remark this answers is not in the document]
MarcelGeo: i am creating own file name in FE ... this is just placeholder.

server/mergin/tests/test_statistics.py
MarcelGeo: yeah, we discussed with TM and we should send it when collect is false.

MarcelGeo: We need to remove every if before call of send statistics or split these things to 2 celeries.

server/mergin/stats/tasks.py
MarcelGeo: then it could be separated celery

web-app/packages/lib/src/common/date_utils.ts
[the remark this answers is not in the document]
varmar05: I meant it also because of some gymnastics in store.ts ...

server/mergin/stats/controller.py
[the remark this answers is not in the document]
MarcelGeo: done

server/mergin/stats/tasks.py
MarcelGeo: done

server/mergin/stats/models.py
```diff
@@ -19,3 +20,14 @@ def __init__(self, service_id: str = None):
...
+
+class MerginStatistics(db.Model):
+    """Information about deployment"""
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    created_at = db.Column(
+        db.DateTime, index=True, nullable=False, server_default="now()"
+    )
+    # data with statistics
+    data = db.Column(db.JSON)
```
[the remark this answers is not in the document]
MarcelGeo: done

server/mergin/stats/tasks.py
MarcelGeo: done as separated celery

web-app/packages/lib/src/common/date_utils.ts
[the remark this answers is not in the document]
MarcelGeo: It's about data coming directly from primevue Calendar picker. Standard date library is enough to do this conversions.

server/mergin/stats/models.py
```diff
@@ -2,13 +2,30 @@
...
+from typing import Optional
 import uuid
 from sqlalchemy.dialects.postgresql import UUID
 from datetime import datetime, timezone
 
 from ..app import db
 
 
+@dataclass
+class ServerCallhomeData:
```
varmar05, line 15: I'd probably split this one into call home data and server_usage (the former extending the later). We can then save only server usage to db as it does not make sense to save deployment variables.

server/mergin/stats/tasks.py
```diff
@@ -2,20 +2,67 @@
...
+
+@celery.task(ignore_result=True)
+def save_statistics():
+    """Save statistics about usage."""
+    to_date = datetime.now(timezone.utc)
+    from_date = datetime.now(timezone.utc) - timedelta(hours=12)
+    # debounce to avoid sending multiple times in short after restart of server
+    last_stat = MerginStatistics.query.filter(
+        MerginStatistics.created_at > from_date,
+        MerginStatistics.created_at < to_date,
```
varmar05, line 54: do we need this?

server/mergin/stats/tasks.py
```diff
@@ -2,20 +2,67 @@
...
+    ).first()
+    if last_stat:
+        return
+
+    info = MerginInfo.query.first()
+    data = get_callhome_data(info)
```
varmar05, line 60: see comments about data structure

server/mergin/stats/controller.py
```diff
@@ -62,25 +63,24 @@ def download_report(date_from: str, date_to: str):
...
 
     stats = (
-        db.session.query(MerginStatistics.data)
+        db.session.query(MerginStatistics.created_at, MerginStatistics.data)
         .filter(MerginStatistics.created_at.between(parsed_from, parsed_to))
         .order_by(MerginStatistics.created_at.desc())
         .all()
     )
-    data = [stat.data for stat in stats]
+    createdColumn = "created_at"
```
varmar05, line 71: use snake_case

web-app/packages/lib/src/common/date_utils.ts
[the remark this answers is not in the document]
MarcelGeo: I moved to already used date-fns for better life :D

server/mergin/stats/models.py
MarcelGeo: ok, but the target is to store everything we are sending to statistics endpoint to database. As we do not know about servers options when somebody will send us this report.

server/mergin/stats/tasks.py
MarcelGeo: we are checking this to prevent send it more times after server restarts

server/mergin/stats/controller.py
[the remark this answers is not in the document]
varmar05: ah ok, it is clearer now!

server/mergin/stats/models.py
varmar05: fair enough, let's keep it as is

varmar05 (approved):

PR #378, opened 2025-02-19
Added abilty to store breadcrumbs in store, not just from router

MarcelGeo:
- there are bugs in vue-router and query strings and meta


harminius (approved):
Looks good

PR #379, opened 2025-02-20
Drop error details index gh#2776

harminius:
Fixes: [#2776](https://github.com/MerginMaps/server-private/issues/2776)

On production we'll use the following:
`DROP INDEX CONCURRENTLY IF EXISTS ix_sync_failures_history_error_details;`
to avoid downtime, instead.

### DIFF
--- a/server/migrations/community/5ad13be6f7ef_drop_error_details_index.py
+++ b/server/migrations/community/5ad13be6f7ef_drop_error_details_index.py
@@ -1,0 +1,33 @@
+"""Drop error details index
+
+Revision ID: 5ad13be6f7ef
+Revises: ba5051218de4
+Create Date: 2025-02-20 07:52:36.670158
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from alembic import context
+
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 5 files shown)


server/migrations/community/5ad13be6f7ef_drop_error_details_index.py
```diff
@@ -0,0 +1,38 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    conn.execute(
+        sa.text(
+            """
+                DROP INDEX IF EXISTS ix_sync_failures_history_error_details;
```
varmar05, line 26: Do not we need to drop this also from models.py definition?

harminius: :facepalm: I had it in my mind and then I disappeared. Fixed.

varmar05: no problem

varmar05 (approved):

MarcelGeo (approved):

PR #384, opened 2025-02-26
Allow admin@example.com also in tests

MarcelGeo:


harminius (commented):
Not strictly connected to this PR, but to the generate_username method..
I'm unsure about this:
image

Although it looks good and I'm happy with the idea behind that I think that doesn't work correctly because "{username}0" could already exist.
Either someone could pick that username manually or there could be `support@microsoft.com` and `support@openai.com` will fail.

harminius (approved):

PR #389, opened 2025-03-03
Add db rollback on gevent timeout error

varmar05:
Handle gevent exception when it occurs during db.session. Normally exception is propagated without any issues, however in case we need to do some cleanups (in `finally` clause) which involve db.session we need a special handling:
- a rollback to unblock cleanup
- reraise exception to be caught by middleware

### DIFF
--- a/server/mergin/tests/test_middleware.py
+++ b/server/mergin/tests/test_middleware.py
@@ -69,3 +69,5 @@ def test_catch_timeout():
     # in case of missing rollback sqlalchemy would raise error
     with pytest.raises(sqlalchemy.exc.PendingRollbackError):
         application.test_client().get("/unhandled")
+    
+    db.session.rollback()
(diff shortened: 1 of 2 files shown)


server/mergin/tests/test_middleware.py
```diff
@@ -31,3 +33,41 @@ def ping():
...
+    application.add_url_rule("/unhandled", "unhandled", unhandled)
+    application.add_url_rule("/timeout", "timeout", timeout)
+    app_context = application.app_context()
+    app_context.push()
+
+    assert application.test_client().get("/timeout").status_code == 504
+
+    # in case of missing rollback sqlalchemy would raise error
+    with pytest.raises(sqlalchemy.exc.PendingRollbackError):
+        application.test_client().get("/unhandled")
```
MarcelGeo, line 71: some assert here, or?

MarcelGeo (changes requested):
Just one thing in tests.

We do not have any other places in code? If not , ok :)

MarcelGeo (approved):

PR #391, opened 2025-03-04
Use the correct method when sharing with new user

harminius:
Fixes: https://github.com/MerginMaps/server-private/issues/2799

Problem:
For CE we use a new API to list project collaborators. We forgot to call it after the invitation is sent.

Solution:
Replace the legacy `getProjectAccess` method with the new `getProjectCollaborators`.


harminius:
> Just double check if something in ee/cloud will be not affected.

Double checked and it works in ee.

PR #394, opened 2025-03-05
Fix generate_username to be case insensitive

varmar05:
For autogenerated usernames make sure we do case insensitive lookup in db.


MarcelGeo (approved):

harminius (approved):
:rocket:

PR #398, opened 2025-03-07
Update version to 2025.2.2

harminius:


MarcelGeo (commented):
Change it also in setup.py @harminius

PR #400, opened 2025-03-07
Add map route when appropriate

harminius:
Introduce showMap prop to control displaying Map tab.


web-app/packages/lib/src/modules/project/views/ProjectViewTemplate.vue
```diff
@@ -150,6 +150,7 @@ export default defineComponent({
     namespace: String,
     projectName: String,
     showSettings: Boolean as PropType<boolean>,
+    showMap: Boolean as PropType<boolean>,
```
MarcelGeo, line 153: I think we can control it in private repositories. Do not inject map template in case of some conditions.

MarcelGeo (changes requested):

MarcelGeo:
we do not need it

PR #401, opened 2025-03-10
Publish docker images to dockerhub

varmar05:
Add workflow dispatch action to build and publish public docker images.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,40 @@
+name: Build and publish public images
+on:
+  workflow_dispatch:
+
+env:
+  REGISTRY: docker.io
+  REPOSITORY: lutraconsulting/merginmaps
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,40 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ secrets.DOCKERHUB_LOGIN }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build, tag, and push CE images to Dockerhub
+        env:
+          BUILD_HASH: ${{ github.sha }}
+        run: |
+          IMAGE_TAG=${{ github.ref_name }}
```
MarcelGeo, line 25: i think we can define this in env above.

MarcelGeo (approved):

PR #406, opened 2025-03-12
Added severity option to TipMessage

MarcelGeo:
Added :heavy_plus_sign: 
- new color negative-light-color
- add message option to confirm dialog to show TipMessage with props
- make confirm dialog more dense

Fixed :hammer: :
- remove tip message from DeleteAccountDialog


harminius (commented):
Are you sure it is clear that a user knows what value they should type in?

harminius (approved):
Nice

Issue #408, opened 2025-03-13
Extend server check for permissions


fernandinand:
Extend `commands.py` server check for permissions related issues.

PR #409, opened 2025-03-13
Add permissions test for server check command #408

fernandinand:


server/mergin/commands.py
```diff
@@ -117,6 +127,12 @@ def _check_server():  # pylint: disable=W0612
         else:
             click.secho("Database initialized properly", fg="green")
 
+        if edition == "Community Edition":
```
varmar05, line 130: we usually do not do this kind of if-else branches as it is difficult to maintain
usually in other versions we have overloaded/extended functions which do the right job

we can only check data dir for sync as it is the most critical and common here

server/mergin/commands.py
```diff
@@ -82,6 +82,16 @@ def _send_email(email: str):
                 f"Error sending email: {e}",
             )
 
+    def _check_permissions(path):
+        import os
```
varmar05, line 86: any reason for local import?

server/mergin/commands.py
```diff
@@ -82,6 +82,16 @@ def _send_email(email: str):
                 f"Error sending email: {e}",
             )
 
+    def _check_permissions(path):
```
varmar05, line 85: It might be useful to use this in standalone command with path as argument

server/mergin/commands.py
```diff
@@ -117,6 +127,12 @@ def _check_server():  # pylint: disable=W0612
         else:
             click.secho("Database initialized properly", fg="green")
 
+        if edition == "Community Edition":
+            _check_permissions('/data')
```
varmar05, line 131: this comes from config / env variable

server/mergin/commands.py
```diff
@@ -117,6 +127,12 @@ def _check_server():  # pylint: disable=W0612
...
+        else:
+            _check_permissions('/data')
+            _check_permissions('/overviews')
```
varmar05, line 134: I would not do this here and move it to maps (maybe as a part of more generic check?), moreover we should not forget that this check would need to be done from celery container not the server one

varmar05 (changes requested):

server/mergin/commands.py
```diff
@@ -1,3 +1,5 @@
+from email.policy import default
```
varmar05, line 1: where is this one used?

server/mergin/commands.py
```diff
@@ -222,3 +224,9 @@ def send_check_email(email: str):  # pylint: disable=W0612
     def check():
         """Check server configuration."""
         _check_server()
+
+    @server.command()
+    @click.option("--path", required=False, default="/data")
```
varmar05, line 229: we should really avoid hardcoded paths - it should be read from config

varmar05 (changes requested):

server/mergin/commands.py
fernandinand: Seems like PyCharm auto thing...removing

server/mergin/commands.py
```diff
@@ -117,6 +128,8 @@ def _check_server():  # pylint: disable=W0612
         else:
             click.secho("Database initialized properly", fg="green")
 
+        _check_permissions("/data")
```
MarcelGeo, line 131: Just final check. Why is this "/data" hardcoded?

MarcelGeo (changes requested):

MarcelGeo (approved):

varmar05 (approved):

server/mergin/app.py
```diff
@@ -189,6 +189,7 @@ def create_app(public_keys: List[str] = None) -> Flask:
     login_manager.init_app(app.app)
     # register auth blueprint
     register_auth(app.app)
+    server_commands(app.app)
```
varmar05, line 192: This was already registered with create_simple_app function, no?

varmar05 (changes requested):

server/mergin/app.py
MarcelGeo: Yeah, but there is not registered some variable if it's in create simple app.

I think it's better to just move it here. Instead copy

varmar05: yea, let's move it here

varmar05 (approved):

PR #413, opened 2025-03-20
Fix access request emails:

MarcelGeo:
- introduce get_email_receivers method which will get users able to get mails about project

### DIFF
--- a/server/mergin/sync/project_handler.py
+++ b/server/mergin/sync/project_handler.py
@@ -1,4 +1,4 @@
-from .models import ProjectRole, ProjectUser
+from .models import ProjectRole, ProjectUser, Project
 from .interfaces import AbstractProjectHandler
 from .permissions import ProjectPermissions
 from sqlalchemy import or_, and_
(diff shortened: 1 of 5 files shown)


server/mergin/sync/project_handler.py
```diff
@@ -1,7 +1,29 @@
...
+            .outerjoin(ProjectUser, ProjectUser.user_id == User.id)
+            .filter(
+                or_(
+                    and_(
+                        ProjectUser.project_id == project.id,
+                        ProjectUser.role == ProjectRole.OWNER.value,
+                    ),
+                    User.is_admin,
+                ),
+                User.verified_email,
```
varmar05, line 25: let's also add filter for active users only

server/mergin/sync/project_handler.py
```diff
@@ -1,7 +1,29 @@
...
+        return (
+            User.query.join(UserProfile)
+            .outerjoin(ProjectUser, ProjectUser.user_id == User.id)
+            .filter(
+                or_(
+                    and_(
+                        ProjectUser.project_id == project.id,
+                        ProjectUser.role == ProjectRole.OWNER.value,
+                    ),
+                    User.is_admin,
```
harminius, line 23: should be the workspace admin, no the super admin?

varmar05 (approved):

PR #416, opened 2025-04-01
Don't allow SSO users update their profile

harminius:
Users that signed in with SSO are not allowed to change their email, name or password.

Frontend does not show the button to update the profile. :empty_nest: 

Backend blocks profile update :no_entry: 

image

Unittest only for admin endpoint as I'm not able to login as SSO user (right now)

### DIFF
--- a/server/mergin/auth/app.py
+++ b/server/mergin/auth/app.py
@@ -11,7 +11,7 @@ from sqlalchemy import func
 
 from .commands import add_commands
 from .config import Configuration
-from .models import User, UserProfile
+from .models import User
 
 # signal for other versions to listen to
 user_account_closed = signal("user_account_closed")
@@ -67,6 +67,17 @@ def auth_required(f=None, permissions=None):
                     return "Permission denied.", 403
         return f(*args, **kwargs)
 
+    return wrapped_func
+
+
+def edit_profile_enabled(f):
+    from .controller import CANNOT_EDIT_PROFILE_MSG
+
+    @functools.wraps(f)
+    def wrapped_func(*args, **kwargs):
+        if not current_user.can_edit_profile:
+            return CANNOT_EDIT_PROFILE_MSG, 403
+        return f(*args, **kwargs)
     return wrapped_func
 
 
--- a/server/mergin/auth/app.py
+++ b/server/mergin/auth/app.py
@@ -78,6 +78,7 @@ def edit_profile_enabled(f):
         if not current_user.can_edit_profile:
             return CANNOT_EDIT_PROFILE_MSG, 403
         return f(*args, **kwargs)
+
     return wrapped_func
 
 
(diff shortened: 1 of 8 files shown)


server/mergin/auth/models.py
```diff
@@ -41,8 +41,8 @@ class User(db.Model):
         db.Index("ix_user_email", func.lower(email), unique=True),
     )
 
-    def __init__(self, username, email, passwd, is_admin=False):
-        self.username = username
+    def __init__(self, username, email, passwd=None, is_admin=False):
+        self.username = username or self.generate_username(email)
```
varmar05, line 45: I would keep username as required and generate it beforehand

server/mergin/auth/controller.py
```diff
@@ -255,6 +255,8 @@ def logout():  # pylint: disable=W0613,W0612
 
 @auth_required
 def change_password():  # pylint: disable=W0613,W0612
+    if not current_user.can_edit_profile:
+        abort(403, "You cannot edit your profile")
```
varmar05, line 259: create a constant with the err message

server/mergin/auth/models.py
```diff
@@ -236,6 +240,12 @@ def create(
         db.session.commit()
         return user
 
+    @property
+    def can_edit_profile(self) -> bool:
+        """Flag if we allow user to edit their email and name"""
+        # False when user is created by SSO login
+        return self.passwd is not None
```
varmar05, line 247: maybe redundant but this should be only for active users

varmar05 (changes requested):

server/mergin/auth/app.py
```diff
@@ -70,6 +70,17 @@ def wrapped_func(*args, **kwargs):
     return wrapped_func
 
 
+def edit_profile_enabled(f):
+    from .controller import CANNOT_EDIT_PROFILE_MSG
```
varmar05, line 74: maybe better to move msg here as app is the main file of blueprint

server/mergin/auth/app.py
```diff
@@ -70,6 +70,17 @@ def wrapped_func(*args, **kwargs):
...
+
+    @functools.wraps(f)
+    def wrapped_func(*args, **kwargs):
+        if not current_user.can_edit_profile:
```
varmar05, line 78: check that current user is not none / anonymous

server/mergin/auth/controller.py
```diff
@@ -255,6 +256,7 @@ def logout():  # pylint: disable=W0613,W0612
 
 
 @auth_required
+@edit_profile_enabled
 def change_password():  # pylint: disable=W0613,W0612
     if not current_user.can_edit_profile:
```
varmar05, line 261: rm in favor of decorator

varmar05 (changes requested):

varmar05 (approved):

PR #422, opened 2025-04-10
Add possibility to lock a project

harminius:
Resolves https://github.com/MerginMaps/server-private/issues/2836

- add column to project table
- block push
- do not count to workspace storage and project number

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -88,7 +88,6 @@ class Project(db.Model):
         self.creator = creator
         self.latest_version = 0
         self.public = kwargs.get("public", False)
-        self.locked_until = None
         latest_files = LatestProjectFiles(project=self)
         db.session.add(latest_files)
 
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -16,7 +16,6 @@ from flask_login import current_user
 from pygeodiff import GeoDiff
 from sqlalchemy import text, null, desc, nullslast
 from sqlalchemy.dialects.postgresql import ARRAY, BIGINT, UUID, JSONB, ENUM
-from sqlalchemy.orm.attributes import flag_modified
 from sqlalchemy.types import String
 from sqlalchemy.ext.hybrid import hybrid_property
 from pygeodiff.geodifflib import GeoDiffLibError
@@ -68,6 +67,7 @@ class Project(db.Model):
         db.Integer, db.ForeignKey("user.id"), nullable=True, index=True
     )
     public = db.Column(db.Boolean, default=False, index=True, nullable=False)
+    locked_until = db.Column(db.DateTime, index=True)
 
     creator = db.relationship(
         "User", uselist=False, backref=db.backref("projects"), foreign_keys=[creator_id]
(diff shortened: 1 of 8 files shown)


server/mergin/sync/models.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         self.creator = creator
         self.latest_version = 0
         self.public = kwargs.get("public", False)
+        self.locked_until = None
```
varmar05, line 91: I'd assume server default null

server/mergin/sync/public_api_controller.py
```diff
@@ -785,6 +785,11 @@ def project_push(namespace, project_name):
     changes = request.json["changes"]
     project_permission = current_app.project_handler.get_push_permission(changes)
     project = require_project(namespace, project_name, project_permission)
+    if project.locked_until:
+        abort(
+            423,
+            f"This project is currently locked. You cannot make changes to it until {project.locked_until.strftime('%Y-%m-%d %H:%M UTC')}.",
```
varmar05, line 791: Do we need this time information? Project would be most likely removed anyway.

server/mergin/sync/public_api_controller.py
```diff
@@ -785,6 +785,11 @@ def project_push(namespace, project_name):
     changes = request.json["changes"]
     project_permission = current_app.project_handler.get_push_permission(changes)
     project = require_project(namespace, project_name, project_permission)
+    if project.locked_until:
+        abort(
+            423,
```
varmar05, line 790: not sure we need custom error code but if we do, let's extend api.yaml

server/mergin/tests/test_project_controller.py
```diff
@@ -2621,3 +2621,33 @@ def test_supported_file_upload(client):
         resp.json["detail"]
         == f"Unsupported file type detected: {spoof_name}. Please remove the file or try compressing it into a ZIP file before uploading."
     )
+
+
+def test_locked_project(client, diff_project):
+    """Users cannot push to the locked project. Moreover, it does not count to the storage and project count67"""
```
varmar05, line 2627: typo

varmar05 (changes requested):
missing alembic migration?

server/mergin/sync/public_api_controller.py
```diff
@@ -787,8 +787,8 @@ def project_push(namespace, project_name):
     project = require_project(namespace, project_name, project_permission)
     if project.locked_until:
         abort(
-            423,
-            f"This project is currently locked. You cannot make changes to it until {project.locked_until.strftime('%Y-%m-%d %H:%M UTC')}.",
+            400,
+            f"This project is currently locked and you cannot make changes to it.",
```
varmar05, line 791: Maybe liaise with Tomas about error code, mobile clients might benefit from a custom one?

varmar05 (approved):

server/mergin/sync/public_api_controller.py
tomasMizera: Yeah, please send it with the new error codes structure (note: we do not care about the specific http response code as long as it is 4xx)

harminius: https://github.com/MerginMaps/server/pull/422/commits/6e07388a44cc63fe249f8f63b381aa484b856b5f

PR #426, opened 2025-04-11
Initial version for zip download rework

varmar05:
Rewrite download zip functionality so it creates a background task for zip construction first and only serves content when ready.

So far this is only prototype with some basic workflow
```
merginmaps-server   | [ACCESS] 172.18.0.1 GET /v1/project/download/mergin/test format=zip HTTP/1.1 202 50 http://localhost:8080/projects/mergin/test/tree "Mozilla/5.0 (X11; Linux x86_64; rv:137.0) Gecko/20100101 Firefox/137.0" 104532 <8>
celery-worker       | [2025-04-11 11:55:51,044: INFO/ForkPoolWorker-1] Task mergin.sync.tasks.create_project_version_zip[d9d90279-ae16-4afd-bd10-2d79cc719df6] succeeded in 0.13913617400248768s: None
```

TODO:
- ~~rename endpoint (make it private)~~
- ~~fix tests~~
- ~~remove unused download functions and libraries, clean up~~
- ~~frontend~~
- optionally move locking from FS to redis [NOT PLANNED]
- ~~update nginx config, make sure buffering settings are correct [:question: ]~~
- ~~make sure, that cleanup of partial is ok - add to controller safe check if .partial is not update more than 30minutes.~~

**Dashboard**
* introducing polling for download zip file in following way:
* sending HEAD request to the server until 200 is not returned, if 200 is returned -> download file
* added loading indicator and disable dowloading
image

* added maximum of 100 retries for download

:hammer: 
* do not download files to js blob, use traditional anchor click (sending path to filesaver.saveFile() method)


server/mergin/sync/private_api.yaml
```diff
@@ -383,6 +383,31 @@ paths:
...
+      parameters:
+        - $ref: "#/components/parameters/ProjectId"
+        - name: version
+          in: query
+          description: Particular version to download
+          required: false
+          schema:
+            $ref: "#/components/schemas/VersionName"
+      responses:
+        "202":
```
MarcelGeo, line 402: there is also 200 or something from send_file.

server/mergin/tests/test_private_project_api.py
```diff
@@ -418,3 +420,50 @@ def test_admin_project_list(client):
     p.delete()
     resp = client.get("/app/admin/projects?page=1&per_page=15&like=mergin")
     assert len(resp.json["items"]) == 14
+
+
+test_download_proj_data = [
+    (None, 202),
```
varmar05, line 426: let's add also test for 200

server/mergin/sync/utils.py
```diff
@@ -546,3 +547,21 @@ def is_supported_type(filepath) -> bool:
 def get_mimetype(filepath: str) -> str:
     """Identifies file types by checking their headers"""
     return magic.from_file(filepath, mime=True)
+
+
+def get_x_accel_uri(*url_parts):
+    """
+    Get the accell uri for the given url parts
```
varmar05, line 554: Would be good to write some example here as it seems bit confusing. 

How about this way:

- pass absolute path to file on disk (e.g. /data/projectA/file.txt)
- verify path starts with LOCAL_PROJECTS and then replace with '/download' and do it posix way

server/mergin/sync/utils.py
```diff
@@ -546,3 +547,34 @@ def is_supported_type(filepath) -> bool:
...
+        str: A URI string starting with "/download", followed by the joined
+             and adjusted path based on the provided URL parts.
+    Example:
+        Assuming `current_app.config["LOCAL_PROJECTS"]` is set to
+        "/home":
+        >>> get_x_accel_uri("/home", "example", "file.txt")
+        '/download/example/file.txt'
+    """
+    download_accell_uri = "/download"
+    if not url_parts:
```
varmar05, line 570: why we need parts and not abs path?

server/mergin/sync/utils.py
```diff
@@ -546,3 +547,34 @@ def is_supported_type(filepath) -> bool:
...
+        return download_accell_uri
+
+    local_projects = current_app.config.get("LOCAL_PROJECTS")
+    print(local_projects)
```
varmar05, line 574: rm

varmar05:
Could you guys also double check we have disabled response buffers on proxy for this endpoint?

server/mergin/sync/utils.py
MarcelGeo: You do not need use os.path.join in code.

MarcelGeo:
> Could you guys also double check we have disabled response buffers on proxy for this endpoint?

I updated no buffer endpoint in nginx.conf. See https://github.com/MerginMaps/server/pull/426/commits/3f31b3fbc1890b7684444553883195fc5ff79c98. There are :snowboarder: questions about that:

- there was trailing slash on the end of directive, I think It was not working before (could we check our deployments?)
- there was enabled / trying to enable to X-Accel-Buffering with true. By default it's "yes" - it's not causing bad UX with existing download solution?
- I disabled buffering, but after testing on staging, we could introduce changes

server/mergin/sync/private_api.yaml
harminius: https://github.com/MerginMaps/server/pull/426/commits/ca448c6e1cbeef2dd852fbdebd6859328238362b ok?

MarcelGeo:
@varmar05 introduced configurable buffering for x accel in download zip.

MarcelGeo (approved):

PR #433, opened 2025-04-29
Fix import

harminius:
The broken import was in the branch that checks if there is an inactive partial zip. 
So it failed each time when the response was not immediate.

### DIFF
--- a/server/mergin/sync/private_api_controller.py
+++ b/server/mergin/sync/private_api_controller.py
@@ -45,7 +45,6 @@ from .storages.disk import move_to_tmp
 from .utils import get_x_accel_uri
 
 project_access_granted = signal("project_access_granted")
-PARTIAL_ZIP_EXPIRATION = 300  # seconds
 
 
 @auth_required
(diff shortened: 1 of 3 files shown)


server/mergin/sync/private_api_controller.py
```diff
@@ -366,7 +366,7 @@ def download_project(id: str, version=None):  # noqa: E501 # pylint: disable=W06
     # to be safe we are not in vicious circle remove inactive partial zip
     if os.path.exists(temp_zip_path) and datetime.fromtimestamp(
         os.path.getmtime(temp_zip_path)
-    ) < datetime.now(datetime.timezone.utc) - timedelta(
+    ) < datetime.now(timezone.utc) - timedelta(
```
varmar05, line 369: Could you add a test for this to make sure we get it covered?

varmar05 (changes requested):

server/mergin/sync/private_api_controller.py
harminius: here's it is https://github.com/MerginMaps/server/pull/433/commits/0a763efdfb75bc349a4145fbeaf156eb30f8d919

MarcelGeo (approved):

PR #450, opened 2025-05-07
Cleanup of permission utils from un neccessary objects

MarcelGeo:
- remove 4113 from download apis

Just :clamp:


harminius (approved):
Nice

PR #454, opened 2025-05-12
Make tests working

MarcelGeo:
- there was one failing test, it looks randomly. But it was failing everytime api.yaml had same changes. I added 422 as return value for projects download and tests are running :green_circle: . There is nowhere such fail. Any ides? I think it's working now :imp:


varmar05 (approved):
Smells like some dark magic (or pg connection issue :-)). Let's see if it breaks again at some point.

PR #460, opened 2025-05-30
Breacrumbs fix + small app header enhancements

MarcelGeo:
Fixed :hammer: 

- app breadcrumbs component now accepts also function in meta, where are parsed route params
- there was issue with beforeEnter handler, where meta was not provided properly
- switching breadcrumbs in case of files or version details are showing

Screenshot from 2025-05-30 14-34-46

Enhancements:

- add sign in button for public view of projects
- get rid of console error in case of non logged user

### DIFF
--- a/web-app/packages/app/src/router.ts
+++ b/web-app/packages/app/src/router.ts
@@ -194,75 +194,82 @@ export const createRouter = (pinia: Pinia) => {
             name: ProjectRouteName.ProjectTree,
             component: FileBrowserView,
             props: true,
-            meta: { public: true }
+            meta: {
+              public: true,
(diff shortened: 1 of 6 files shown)


harminius (changes requested):
I suggest to run it locally.
I cannot make it working 

[Screencast from 02-06-25 11:04:14.webm](https://github.com/user-attachments/assets/0c7829c6-aa79-46cd-8cb0-a9098a38aab3)

MarcelGeo:
> I suggest to run it locally. I cannot make it working
> Screencast.from.02-06-25.11.04.14.webm

Yeah, it was running localy by development. Try to `yarn build:libs`

PR #464, opened 2025-06-11
Allow concurrent non-blocking uploads

varmar05:
Introduce blocking and non-blocking uploads while we allow for multiple concurrent non-blocking uploads to be created.
On DB level this constraint is achieved with partial index on blocking uploads.

Every upload creates subsequent project version so we do not specify which version should be created.
There is integrity check before creating project version in case of race conditions - clients should retry on 409 which would then finalize transaction with bumped project version.

### DIFF
--- a/server/mergin/sync/public_api_controller.py
+++ b/server/mergin/sync/public_api_controller.py
@@ -10,8 +10,8 @@ import logging
 from dataclasses import asdict
 from typing import Dict
 from urllib.parse import quote
-import uuid
 from datetime import datetime
+from marshmallow import ValidationError
 
 import gevent
 import psycopg2
@@ -50,11 +50,13 @@ from .models import (
     ProjectRole,
 )
 from .files import (
-    UploadChanges,
     ChangesSchema,
-    UploadFileSchema,
+    ProjectDiffFile,
+    ProjectFile,
+    ProjectFileChange,
     ProjectFileSchema,
-    FileSchema,
+    files_changes_from_upload,
+    mergin_secure_filename,
 )
 from .schemas import (
     ProjectSchema,
@@ -239,15 +241,24 @@ def add_project(namespace):  # noqa: E501
                 .first_or_404()
             )
             version_name = 1
-            files = UploadFileSchema(context={"version": 1}, many=True).load(
-                FileSchema(exclude=("location",), many=True).dump(template.files)
-            )
-            changes = UploadChanges(added=files, updated=[], removed=[])
+            file_changes = []
+            for file in template.files:
+                file_changes.append(
+                    ProjectFileChange(
+                        file.path,
+                        file.checksum,
+                        file.size,
+                        diff=None,
+                        mtime=None,
+                        location=os.path.join("v1", mergin_secure_filename(file.path)),
+                        change=PushChangeType.CREATE,
+                    )
+                )
 
         else:
             template = None
             version_name = 0
-            changes = UploadChanges(added=[], updated=[], removed=[])
+            file_changes = []
 
         try:
             p.storage.initialize(template_project=template)
@@ -258,7 +269,7 @@ def add_project(namespace):  # noqa: E501
             p,
             version_name,
             current_user.id,
-            changes,
+            file_changes,
             get_ip(request),
             get_user_agent(request),
             get_device_id(request),
@@ -758,13 +769,15 @@ def project_push(namespace, project_name):
     if not pv and version != 0:
         abort(400, "First push should be with v0")
 
-    if all(len(changes[key]) == 0 for key in changes.keys()):
-        abort(400, "No changes")
-
-    upload_changes = ChangesSchema(context={"version": pv.name + 1}).load(changes)
+    try:
+        ChangesSchema().validate(changes)
+        upload_changes = ChangesSchema().dump(changes)
+    except ValidationError as err:
+        msg = err.messages[0] if type(err.messages) == list else "Invalid input data"
+        abort(400, msg)
 
     # reject upload early if there is another blocking upload already running
-    if Upload.is_blocking(upload_changes):
+    if upload_changes["is_blocking"]:
         pending_upload = Upload.query.filter_by(
             project_id=project.id, blocking=True
         ).first()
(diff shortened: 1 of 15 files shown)


server/mergin/tests/test_project_controller.py
```diff
@@ -1020,6 +1021,12 @@ def _get_changes_with_diff_0_size(project_dir):
     return changes
 
 
+def _get_non_blocking_changes(project_dir):
```
harminius: is it used?

harminius (approved):
Looks good :rocket:

server/mergin/tests/test_project_controller.py
varmar05: probably just something I intended but did not use at the end, I will remove it

server/mergin/sync/public_api_controller.py
```diff
@@ -752,31 +752,36 @@ def project_push(namespace, project_name):
...
-    pending_upload = Upload.query.filter_by(
-        project_id=project.id, version=version
-    ).first()
-    if pending_upload and pending_upload.is_active():
-        abort(400, "Another process is running. Please try later.")
+    upload_changes = ChangesSchema(context={"version": pv.name + 1}).load(changes)
 
-    upload_changes = ChangesSchema(context={"version": version + 1}).load(changes)
+    # reject upload early if there is another blocking upload already running
+    if Upload.is_blocking(upload_changes):
```
MarcelGeo: Do we want to make something like

```python
if upload_changes.is_blocking:
  pass
```
Move is_blocking method to `ChangeSchema` as it's just looping trough self. attributes. I think we can then run changes.is_blocking in Upload contructor.

2. Upload is already having blocking attribute and it looks like is_blocking is getter from DB.

server/mergin/sync/public_api_controller.py
```diff
@@ -752,31 +752,36 @@ def project_push(namespace, project_name):
...
     if all(len(changes[key]) == 0 for key in changes.keys()):
         abort(400, "No changes")
 
-    # reject upload early if there is another one already running
-    pending_upload = Upload.query.filter_by(
-        project_id=project.id, version=version
-    ).first()
-    if pending_upload and pending_upload.is_active():
-        abort(400, "Another process is running. Please try later.")
+    upload_changes = ChangesSchema(context={"version": pv.name + 1}).load(changes)
```
MarcelGeo: Just question: why we need version to ChangeSchema?

server/mergin/sync/tasks.py
```diff
@@ -155,3 +155,12 @@ def remove_projects_archives():
                 os.remove(path)
             except OSError as e:
                 logging.error(f"Unable to remove {path}: {str(e)}")
+
+
+@celery.task
+def remove_stale_project_uploads(project_id: str):
```
MarcelGeo, line 161: how was this handled before?

server/mergin/sync/public_api_controller.py
```diff
@@ -938,9 +943,9 @@ def chunk_upload(transaction_id, chunk_id):
     """
     upload, upload_dir = get_upload(transaction_id)
     request.view_args["project"] = upload.project
-    upload_changes = ChangesSchema(context={"version": upload.version + 1}).load(
-        upload.changes
-    )
+    upload_changes = ChangesSchema(
+        context={"version": upload.project.latest_version + 1}
```
MarcelGeo: this could be problem if there is different number as in push_finish, no?

server/mergin/sync/public_api_controller.py
```diff
@@ -1135,10 +1135,14 @@ def push_finish(transaction_id):
             f"Failed to finish push for project: {project.id}, project version: {v_next_version}, "
             f"transaction id: {transaction_id}.: {str(err)}"
         )
+        if os.path.exists(target_dir):
```
MarcelGeo, line 1146: why it was not here before?

server/mergin/sync/public_api_controller.py
```diff
@@ -984,13 +989,15 @@ def push_finish(transaction_id):
...
-    changes = ChangesSchema(context={"version": upload.version + 1}).load(
-        upload.changes
-    )
     project = upload.project
     if project.locked_until:
         abort(make_response(jsonify(ProjectLocked().to_dict()), 422))
+
     project_path = get_project_path(project)
+    next_version = project.latest_version + 1
+    v_next_version = ProjectVersion.to_v_name(next_version)
```
MarcelGeo, line 985: this should go down for ProjectVersion creation. Or?

server/mergin/sync/public_api_controller.py
```diff
@@ -752,31 +752,36 @@ def project_push(namespace, project_name):
...
         abort(404)
 
-    # fixme use get_latest
-    pv = ProjectVersion.query.filter_by(
-        project_id=project.id, name=project.latest_version
-    ).first()
-    if pv and pv.name != version:
-        abort(400, "Version mismatch")
+    pv = project.get_latest_version()
+    if pv and pv.name < version:
```
MarcelGeo, line 767: Do we really need this version dance? version number is now for me assgined just after everything is fine. Should we get rid of this version param?

MarcelGeo (changes requested):
Looks good. Just questions and some minor tweaks.

1. In general, I do not understand, why is there this version number handling in push start. 

2. Why are we guessing these changes? There was deal, that clients will divide transactions, no?

server/mergin/sync/public_api_controller.py
varmar05: before there was this strict requirement telling client to always pull - now this would not work if we allow for concurrent uploads but the new condition will still protect push to broken project (e.g. project was reverted on server and no client should be able to push to it)

server/mergin/sync/public_api_controller.py
varmar05: this seems to assign location to files (where will they actually be stored) but I think this schema needs refactoring - I just did not want to touch it during this major change

server/mergin/sync/public_api_controller.py
varmar05: yes, we can do it that way

server/mergin/sync/public_api_controller.py
varmar05: I do not think so, at this point we only save chunks into working dir, so this is just confusing due to wrong interface - see my comment about refactoring

server/mergin/sync/public_api_controller.py
varmar05: it is also in some error msg

server/mergin/sync/public_api_controller.py
varmar05: It could have been, but before there was auto clean up on `vx` folder on next push but now we cannot do it this way as sole existence of `vx` folder means that someone else might be pushing

varmar05:
> Looks good. Just questions and some minor tweaks.
> 
>     1. In general, I do not understand, why is there this version number handling in push start.
> 
>     2. Why are we guessing these changes? There was deal, that clients will divide transactions, no? It's porbably because we do not want to break old apis?

1. It is mostly just for some basic consistency checks, we _could_ try to get rid of it
2. Sorry, not sure what you mean by guessing changes. Yes, (new) clients will send multiple transactions and they should not take care about version which will be created. It is server which evaluates both - the nature of push attempt (blocking or not) and next version (head + 1) on successful push.

server/mergin/sync/tasks.py
varmar05: it did not have to be as all stale uploads were removed on next push attempts

server/mergin/sync/public_api_controller.py
MarcelGeo: we could remove it in next iteration ... with freeze

server/mergin/sync/public_api_controller.py
MarcelGeo: ok, moving to refactor.

MarcelGeo (approved):

PR #468, opened 2025-06-17
Refactor upload changes handling

varmar05:
- move relevant validations into upload schema
- use upload schema only in push init while project file object is used for actual file manipulations 
- remove version context and redundant dataclasses for upload schema

### DIFF
--- a/server/mergin/sync/files.py
+++ b/server/mergin/sync/files.py
@@ -3,12 +3,27 @@
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
 import datetime
 import os
+import uuid
 from dataclasses import dataclass
-from typing import Optional, List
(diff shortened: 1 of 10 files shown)


server/mergin/sync/files.py
```diff
@@ -24,94 +39,181 @@ def mergin_secure_filename(filename: str) -> str:
...
+        return bool(
+            len(obj.get("updated", []))
+            or len(obj.get("removed", []))
+            or any(
+                is_qgis(f["path"]) or is_versioned_file(f["path"])
+                for f in obj.get("added", [])
+            )
+        )
+
+    @post_dump
```
MarcelGeo, line 186: We do not want to remove files before loading to schema? We then probably do not need call dump in several places and just work with schema?

varmar05: We probably could filter out blacklisted files earlier but actual reason for calling dump is that we add custom field (`is_blocking`) which is not available on load

PR #473, opened 2025-07-01
Increase max size for diagnostic logs to 10MB

MarcelGeo:


deployment/enterprise/.env.template
```diff
@@ -59,10 +55,20 @@ DB_PORT=5432
 
 # auth related
 
+#USER_SELF_REGISTRATION=True
+
+#SECRET_KEY=NODEFAULT
+SECRET_KEY=fix-me
```
varmar05: let's use `fixme`  everywhere for easier search

MarcelGeo: this was here before :D ok

varmar05 (approved):

PR #475, opened 2025-07-04
Custom bearer token expiration validation

varmar05:
Previously we validated all access tokens with the same expiration value but since SSO we have two types of access token with different expiration period. Therefore, we use `expire` field to validate if token is still valid (instead of built-in max_age validation).

### DIFF
--- a/server/mergin/auth/bearer.py
+++ b/server/mergin/auth/bearer.py
@@ -3,17 +3,28 @@
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
 
 import hashlib
+from datetime import datetime, timezone
 from itsdangerous import URLSafeTimedSerializer
+from itsdangerous.exc import SignatureExpired, BadSignature
 from flask.sessions import TaggedJSONSerializer
 
 
-def decode_token(secret_key, salt, token, max_age=None):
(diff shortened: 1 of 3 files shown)


server/mergin/auth/bearer.py
```diff
@@ -3,17 +3,28 @@
...
     signer_kwargs = {"key_derivation": "hmac", "digest_method": hashlib.sha1}
     s = URLSafeTimedSerializer(
         secret_key, salt=salt, serializer=serializer, signer_kwargs=signer_kwargs
     )
-    return s.loads(token, max_age=max_age)
+    token_data = s.loads(token)
+    try:
+        expire = datetime.fromisoformat(token_data.get("expire"))
+    except (ValueError, TypeError):
+        raise BadSignature("Invalid token")
```
tomasMizera, line 22: Should we consider this as critical failure and log it so we can find it in logs in case something like this would happen? 🤔

server/mergin/auth/bearer.py
```diff
@@ -3,17 +3,28 @@
...
+
+    if expire < datetime.now(timezone.utc):
```
tomasMizera, line 24: When we build the token, we use `pytz.utc`, here when decoding we use `timezone.utc` -- I guess that is allright, right @varmar05?

PR #479, opened 2025-07-15
Fix push finish on racing conditions

varmar05:
Server determines the next version on push finish just before moving files which were already prepared (including geodiff acttions). This should address the issue when during push finish another version is already created.

Workflow in push finish is changed that all hard work is done **before** final version check.

Example:
- on push finish server is currently on v4
- during chunks merge someone else created v5 version
- before commit to db server would bump to v6 for current push finish and move already created files to correct destination

### DIFF
--- a/server/mergin/sync/files.py
+++ b/server/mergin/sync/files.py
@@ -80,17 +80,18 @@ class ProjectFileChange(ProjectFile):
     change: PushChangeType
 
 
-def files_changes_from_upload(changes: dict, version: int) -> List["ProjectFileChange"]:
+def files_changes_from_upload(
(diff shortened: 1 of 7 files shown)


harminius:
When running concurrent sync tests from plugin I still keep hitting the db constraint when committing the same version (see client B log). Could we reintroduce the `409` error code so that client can retry?

[B_client-log.txt](https://github.com/user-attachments/files/21232027/B_client-log.txt)
[A_client-log.txt](https://github.com/user-attachments/files/21232028/A_client-log.txt)

MarcelGeo (approved):

PR #482, opened 2025-07-17
Introduce file diff table

varmar05:
Expand diff json column from file_history into new separate table. At the moment this is just DB refactoring but on the same time it is a prerequisite for further pull improvements. 

Currently we store metadata associated to gpkg-diff files in new table with rank 0 (aka. elementary diff) but the table will host also merged diffs of higher orders in the future.

Every row in the new table is related to:
 - corresponding file change in file_history
 - corresponding basefile (again in file_history) which will be used when diff file is applied

### DIFF
--- a/server/migrations/community/bd1ec73db389_create_file_diff_table.py
+++ b/server/migrations/community/bd1ec73db389_create_file_diff_table.py
@@ -1,0 +1,140 @@
+"""create file diff table
+
+Revision ID: bd1ec73db389
+Revises: 6cb54659c1de
+Create Date: 2025-07-17 14:17:02.373645
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "bd1ec73db389"
+down_revision = "6cb54659c1de"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "file_diff",
+        sa.Column("id", sa.BigInteger(), autoincrement=True, nullable=False),
+        sa.Column("file_path_id", sa.BigInteger(), nullable=False),
+        sa.Column("basefile_id", sa.BigInteger(), nullable=False),
+        sa.Column("path", sa.String(), nullable=False),
+        sa.Column("rank", sa.Integer(), nullable=False),
+        sa.Column("version", sa.Integer(), nullable=False),
+        sa.Column("location", sa.String(), nullable=True),
+        sa.Column("size", sa.BigInteger(), nullable=False),
+        sa.Column("checksum", sa.String(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["basefile_id"],
+            ["file_history.id"],
+            name=op.f("fk_file_diff_basefile_id_file_history"),
+            ondelete="CASCADE",
+        ),
+        sa.ForeignKeyConstraint(
+            ["file_path_id"],
+            ["project_file_path.id"],
+            name=op.f("fk_file_diff_file_path_id_project_file_path"),
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_file_diff")),
+        sa.UniqueConstraint("file_path_id", "rank", "version", name="unique_diff"),
+    )
+    op.create_index(
+        op.f("ix_file_diff_basefile_id"), "file_diff", ["basefile_id"], unique=False
+    )
+    op.create_index(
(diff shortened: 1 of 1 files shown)


server/migrations/community/bd1ec73db389_create_file_diff_table.py
```diff
@@ -0,0 +1,140 @@
...
+        sa.Column("size", sa.BigInteger(), nullable=False),
+        sa.Column("checksum", sa.String(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["basefile_id"],
+            ["file_history.id"],
+            name=op.f("fk_file_diff_basefile_id_file_history"),
+            ondelete="CASCADE",
+        ),
+        sa.ForeignKeyConstraint(
+            ["file_path_id"],
```
MarcelGeo, line 39: why we need this, if file is stored in "file_history" ? Similar situation with version. Please add here also some revision columns (created_at).

server/migrations/community/bd1ec73db389_create_file_diff_table.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+    op.create_index(op.f("ix_file_diff_path"), "file_diff", ["path"], unique=False)
+    op.create_index(op.f("ix_file_diff_rank"), "file_diff", ["rank"], unique=False)
+    op.create_index(
+        op.f("ix_file_diff_version"), "file_diff", ["version"], unique=False
+    )
+
+    # migrate data
+    conn = op.get_bind()
+    conn.execute(
```
MarcelGeo, line 64: Did u check this on real data?
It looks like it works on my machine db :tada:

server/mergin/sync/models.py
```diff
@@ -822,14 +922,18 @@ def _files_from_start(self):
...
+                pv.created AS mtime,
+                fd.path as diff_path,
+                fd.size as diff_size,
+                fd.checksum as diff_checksum,
+                fd.location as diff_location
             FROM latest_changes ch
             LEFT OUTER JOIN file_history fh ON (fh.file_path_id = ch.id AND fh.project_version_name = ch.version)
             LEFT OUTER JOIN project_file_path fp ON fp.id = fh.file_path_id
             LEFT OUTER JOIN project_version pv ON pv.id = fh.version_id
+            LEFT OUTER JOIN file_diff fd ON fd.file_path_id = fh.file_path_id AND fd.version = fh.project_version_name and fd.rank = 0
```
MarcelGeo, line 936: do we need filter for rank here? I think we can upgrade it later.

server/migrations/community/bd1ec73db389_create_file_diff_table.py
```diff
@@ -0,0 +1,140 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "file_diff",
+        sa.Column("id", sa.BigInteger(), autoincrement=True, nullable=False),
+        sa.Column("file_path_id", sa.BigInteger(), nullable=False),
+        sa.Column("basefile_id", sa.BigInteger(), nullable=False),
```
MarcelGeo, line 25: justt to make sure, this is column which is not related to row update_diff value, but to file where it was created?
(id:1) v1 ---> created geopackage
(id:2) v2 ----> updated with geodiff
basefile_id == 1 ?

server/migrations/community/bd1ec73db389_create_file_diff_table.py
varmar05: We need this to:
- track primary dif (rank=0) to its origin in file_history table (as composite FK with version)
- track to which project file merged diff belongs (as it spans over multiple versions and does not have reference in file_history)

created_at is already in project_version table where these can be referenced, it would be only relevant of merged diff but again, the caching job would be initiated on project version creation

server/migrations/community/bd1ec73db389_create_file_diff_table.py
varmar05: yea, I did migration on staging server - forth and back

unfortunately, production db will need custom script where migration will be parallelized

server/mergin/sync/models.py
varmar05: At this stage there will be only ranks 0 so we won't need that but I thought it would be better if query respected more generic logic. Also it revealed the need for composite indexes.

server/migrations/community/bd1ec73db389_create_file_diff_table.py
varmar05: Exactly. So in principle you can then ask: "Give me all diffs with that basefile which I need for upgrade to version X"

PR #486, opened 2025-08-04
Adjust avatar font size globally

harminius:
Fixes https://github.com/MerginMaps/server-private/issues/3041

Adjust avatar font sizes to align with Figma design

image

propagates correctly to members page:

image

### DIFF
--- a/web-app/packages/lib/src/assets/sass/theme-base/components/misc/_avatar.scss
+++ b/web-app/packages/lib/src/assets/sass/theme-base/components/misc/_avatar.scss
@@ -1,6 +1,7 @@
 .p-avatar {
     background-color: $avatarBg;
     border-radius: $borderRadius;
+    font-size: $avatarFontSize !important;
 
     &.p-avatar-lg {
         width: 3rem;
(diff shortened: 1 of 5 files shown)


web-app/packages/lib/src/assets/sass/theme-base/components/misc/_avatar.scss
```diff
@@ -1,6 +1,7 @@
 .p-avatar {
     background-color: $avatarBg;
     border-radius: $borderRadius;
+    font-size: $avatarFontSize !important;
```
MarcelGeo: Do not rewrite these files.

Use _extenstions.scss to tweak .p-avatar class. With something like 
```css
.p-avatar:not(.p-avatar-lg):not(.p-avatar-xl) {
  font-size: 0.875rem;
}
```

We could then double chek if e.g. profile avatar is ok... as it's -lg

Some styles  which are really defaults are comming from inline loader from primeuix library.

image

MarcelGeo (changes requested):

MarcelGeo (approved):

PR #488, opened 2025-08-05
Add new v2 push endpoint to create a project version

varmar05:
- refactor common logic for push in v1 and v2 API
- feature flag to use v2 push (for clients only)
- dry run option in v2 endpoint to only validate request
- use new error structure for v2 endpoint

### DIFF
--- a/server/mergin/sync/files.py
+++ b/server/mergin/sync/files.py
@@ -204,7 +204,7 @@ class ChangesSchema(ma.Schema):
         if len(set(changes_files)) != len(changes_files):
             raise ValidationError("Not unique changes")
 
-        # check if all .gpkg file are valid
+        # check if all files are valid
         for file in data["added"] + data["updated"]:
             file_path = file["path"]
             if is_versioned_file(file_path) and file["size"] == 0:
--- a/server/mergin/sync/public_api_v2.yaml
+++ b/server/mergin/sync/public_api_v2.yaml
@@ -320,7 +320,9 @@ paths:
           content:
             application/problem+json:
               schema:
-                $ref: "#/components/schemas/ProjectVersionMismatch"
+                anyOf:
+                  - $ref: "#/components/schemas/ProjectVersionExists"
+                  - $ref: "#/components/schemas/AnotherUploadRunning"
         "422":
           description: Request could not be processed by server
           content:
@@ -330,8 +332,14 @@ paths:
                   - $ref: "#/components/schemas/UploadError"
                   - $ref: "#/components/schemas/TrialExpired"
                   - $ref: "#/components/schemas/StorageLimitHit"
-                  - $ref: "#/components/schemas/ProjectLocked"
                   - $ref: "#/components/schemas/DataSyncError"
+        "423":
+          description: Project is locked for any upload       
+          content:
+              application/problem+json:
+                schema:
+                  $ref: "#/components/schemas/ProjectLocked"
+
       x-openapi-router-controller: mergin.sync.public_api_v2_controller
 components:
   responses:
@@ -397,7 +405,7 @@ components:
       example:
         code: ProjectLocked
         detail: The project is currently locked and you cannot make changes to it (ProjectLocked)
-    ProjectVersionMismatch:
+    ProjectVersionExists:
       allOf:
         - $ref: '#/components/schemas/CustomError'
       type: object
@@ -407,8 +415,20 @@ components:
         server_version:
           type: string
       example:
-        code: ProjectVersionMismatch
-        detail: Project version mismatch (ProjectVersionMismatch)
+        code: ProjectVersionExists
+        detail: Project version mismatch (ProjectVersionExists)
+    AnotherUploadRunning:
+      allOf:
+        - $ref: '#/components/schemas/CustomError'
+      type: object
+      properties:
+        client_version:
+          type: string
+        server_version:
+          type: string
+      example:
+        code: AnotherUploadRunning
+        detail: Another process is running (AnotherUploadRunning)
     DataSyncError:
       allOf:
         - $ref: '#/components/schemas/CustomError'
--- a/server/mergin/sync/public_api_v2_controller.py
+++ b/server/mergin/sync/public_api_v2_controller.py
@@ -9,7 +9,7 @@ import os
 import psycopg2
 from connexion import NoContent, request
 from datetime import datetime, timedelta, timezone
-from flask import abort, jsonify, current_app, make_response
+from flask import abort, jsonify, current_app
 from flask_login import current_user
 from marshmallow import ValidationError
 from psycopg2 import IntegrityError
@@ -18,10 +18,11 @@ from ..app import db
 from ..auth import auth_required
 from ..auth.models import User
 from .errors import (
+    AnotherUploadRunning,
     BigChunkError,
     DataSyncError,
     ProjectLocked,
-    ProjectVersionMismatch,
+    ProjectVersionExists,
     StorageLimitHit,
     UploadError,
 )
@@ -170,7 +171,7 @@ def create_project_version(id):
     request.view_args["project"] = project
 
     if project.locked_until:
-        return ProjectLocked().response(422)
+        return ProjectLocked().response(423)
 
     next_version = project.next_version()
     v_next_version = ProjectVersion.to_v_name(next_version)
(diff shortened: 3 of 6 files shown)


server/mergin/sync/public_api_v2.yaml
```diff
@@ -244,6 +325,74 @@ components:
         format: uuid
         pattern: \b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b
   schemas:
+    # Errors
+    CustomError:
```
MarcelGeo, line 372: Something like ResponseError?

server/mergin/sync/public_api_v2.yaml
```diff
@@ -244,6 +325,74 @@ components:
...
+      allOf:
+        - $ref: '#/components/schemas/CustomError'
+      type: object
+      properties:
+        client_version:
+          type: string
+        server_version:
+          type: string
+      example:
+        code: ProjectVersionMismatch
```
MarcelGeo: Something like ProjectVersionExists ?

server/mergin/sync/config.py
```diff
@@ -64,10 +64,14 @@ class Configuration(object):
     )
     # in seconds, older unfinished zips are moved to temp
     PARTIAL_ZIP_EXPIRATION = config("PARTIAL_ZIP_EXPIRATION", default=600, cast=int)
+    # whether new push is allowed
+    V2_PUSH_ENABLED = config("V2_PUSH_ENABLED", default=True, cast=bool)
```
MarcelGeo, line 68: do not forget propagate in private

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -136,6 +155,186 @@ def remove_project_collaborator(id, user_id):
...
+    changes = request.json["changes"]
+    project_permission: ProjectPermissions = (
+        current_app.project_handler.get_push_permission(changes)
+    )
+    project = require_project_by_uuid(id, project_permission)
+    # pass full project object to request for later use
+    request.view_args["project"] = project
+
+    if project.locked_until:
+        return ProjectLocked().response(422)
```
MarcelGeo: Just consider to use 423 error code for locked (?)

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -136,6 +155,186 @@ def remove_project_collaborator(id, user_id):
...
+    version_dir = os.path.join(project.storage.project_dir, v_next_version)
+
+    pv = project.get_latest_version()
+    if pv and pv.name != version:
+        return ProjectVersionMismatch(version, pv.name).response(409)
+
+    # reject push if there is another one already running
+    pending_upload = Upload.query.filter_by(project_id=project.id).first()
+    if pending_upload and pending_upload.is_active():
+        return ProjectVersionMismatch(version, next_version).response(409)
```
MarcelGeo: here could be "Another process is running" like in v1?

server/mergin/sync/files.py
```diff
@@ -24,94 +45,181 @@ def mergin_secure_filename(filename: str) -> str:
...
+        ]
+
+        if len(changes_files) == 0:
+            raise ValidationError("No changes")
+
+        # changes' files must be unique
+        if len(set(changes_files)) != len(changes_files):
+            raise ValidationError("Not unique changes")
+
+        # check if all .gpkg file are valid
```
MarcelGeo: This is for all files, no?

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -136,6 +155,186 @@ def remove_project_collaborator(id, user_id):
...
+    # check consistency of changes
+    current_files = set(file.path for file in project.files)
+    added_files = set(file["path"] for file in upload_changes["added"])
+    if added_files and added_files.issubset(current_files):
+        return UploadError(
+            error=f"Add changes contain files which already exist"
+        ).response(422)
+
+    modified_files = set(
+        file["path"] for file in upload_changes["updated"] + upload_changes["removed"]
```
MarcelGeo: could we use some variable for removed, added, updated and use it in code?

server/mergin/sync/files.py
```diff
@@ -24,94 +45,181 @@ def mergin_secure_filename(filename: str) -> str:
...
     class Meta:
         unknown = EXCLUDE
 
-    @post_load
-    def create_obj(self, data, **kwargs):
-        return UploadChanges(**data)
+    @post_dump
+    def remove_blacklisted_files(self, data, **kwargs):
+        """Files which are blacklisted are not allowed to be uploaded and are simple ignored."""
+        for key in ("added", "updated", "removed"):
```
MarcelGeo, line 183: In v1 is here also some sanitization of file name.

server/mergin/sync/models.py
```diff
@@ -1053,6 +1044,141 @@ def clear(self):
...
+        v_next_version = ProjectVersion.to_v_name(self.project.next_version())
+        chunks_map = {
+            f["path"]: f["chunks"]
+            for f in self.changes["added"] + self.changes["updated"]
+        }
+        file_changes = files_changes_from_upload(self.changes, v_next_version)
+        to_remove = [i.path for i in file_changes if i.change == PushChangeType.DELETE]
+        current_files = [f for f in self.project.files if f.path not in to_remove]
+
+        with Toucher(self.lockfile, 5):
```
MarcelGeo, line 1062: Why we changed that from 30 -> 5?

MarcelGeo (commented):
just questions.

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -136,14 +156,194 @@ def remove_project_collaborator(id, user_id):
...
+def create_project_version(id):
+    """Create a new project version from pushed data"""
+    version: int = ProjectVersion.from_v_name(request.json["version"])
+    changes = request.json["changes"]
+    project_permission: ProjectPermissions = (
+        current_app.project_handler.get_push_permission(changes)
+    )
+    project = require_project_by_uuid(id, project_permission)
+    # pass full project object to request for later use
+    request.view_args["project"] = project
```
MarcelGeo, line 171: do we need this?

server/mergin/sync/storages/disk.py
```diff
@@ -245,17 +245,15 @@ def _generator():
         return _generator()
 
     def apply_diff(
-        self, current_file: ProjectFile, upload_file: UploadFile, version: int
+        self, current_file: ProjectFile, changeset: str, patchedfile: str
```
MarcelGeo: patched**_**file

what's changeset? something like geodiff_file_location?

server/mergin/sync/public_api_v2.yaml
varmar05: hm, this is spread all over the place in other yaml files

server/mergin/sync/public_api_v2_controller.py
varmar05: We can do, but probably only in new v2 endpoints, I would not touch the old API

server/mergin/sync/files.py
varmar05: yes, I will update comment

server/mergin/sync/files.py
varmar05: it's been moved here https://github.com/MerginMaps/server/pull/488/files/18abd45c48cc435d58b30242e78f59ac8d9392c0#diff-b386487dac2cf3bb1938dc8d44101adf90a93f3544a9aa793ff5eaaa421c8dfcR89

server/mergin/sync/models.py
varmar05: 30 seemed a lot to me given that requests should be much faster

varmar05: also for other clients waiting it is better to see pending upload being updated more frequently

server/mergin/sync/public_api_v2_controller.py
varmar05: it is used it that catch_error decorator

server/mergin/sync/storages/disk.py
varmar05: yes, changeset is diff file location

varmar05: I can rename it to make it clearer

server/mergin/sync/files.py
MarcelGeo: Just to make sure, this part of code?

```
# all file need to be unique after sanitized
        if f.location in sanitized_files:
            filename, file_extension = os.path.splitext(f.location)
            f.location = filename + f".{str(uuid.uuid4())}" + file_extension
        sanitized_files.append(f.location)
        if f.diff:
            if f.diff.location in sanitized_files:
                filename, file_extension = os.path.splitext(f.diff.location)
                f.diff.location = filename + f".{str(uuid.uuid4())}" + file_extension
            sanitized_files.append(f.diff.location)
```

varmar05: yes, that's the one

PR #489, opened 2025-08-06
Cron job to remove outdated uploaded chunks

varmar05:
Added celery job to regularly remove outdated chunks.

### DIFF
--- a/server/application.py
+++ b/server/application.py
@@ -88,7 +88,7 @@ def setup_periodic_tasks(sender, **kwargs):
         name="remove old project archives",
     ),
     sender.add_periodic_task(
-        crontab(hour="*/12", minute=0),
+        crontab(hour="*/4", minute=0),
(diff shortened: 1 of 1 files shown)


server/application.py
```diff
@@ -85,4 +86,9 @@ def setup_periodic_tasks(sender, **kwargs):
         crontab(hour=3, minute=0),
         remove_projects_archives,
         name="remove old project archives",
+    ),
+    sender.add_periodic_task(
+        crontab(hour="*/12", minute=0),
```
MarcelGeo: are we able to do it configurable?

MarcelGeo (changes requested):

server/application.py
varmar05: let's make it more frequent and leave flexible configuration as separate ticket as it is more generic task.

PR #496, opened 2025-08-15
Fix integrity error handling

varmar05:


varmar05:
> If I tried this with existing Upload entry in the table. v2 project push was not not
> 
> ```sql
> INSERT INTO upload () VALUES ("projectid", version)
> ```
> 
> And tried then `sync_project` in py-client:
> 
>     * pull
> 
>     * push
> 
> 
> Upload entry disappeared and I was able to do push. Not sure If this is good behavior.

Not sure I understand it correctly but if you run raw SQL query (not from application) then you will insert upload record to table but lockfile will not be created, therefore such upload is considered as inactive (invalid) and will be removed by subsequent push attempt.

To have blocking upload you need both, a record in database and recently updated associated lockfile.

MarcelGeo (approved):

PR #497, opened 2025-08-15
Improve download endpoint

harminius:
Fixes https://github.com/MerginMaps/server-private/issues/3061 https://github.com/MerginMaps/server-private/issues/3060

- head request cannot create new celery job
- flask does not remove partial zip
- celery cares of cleaning up abandoned zip partials

### DIFF
--- a/server/mergin/tests/test_private_project_api.py
+++ b/server/mergin/tests/test_private_project_api.py
@@ -460,10 +460,13 @@ def test_download_project(
     project_version = diff_project.get_latest_version()
     if zipfile:
         zip_path = Path(project_version.zip_path)
+        if zip_path.parent.exists():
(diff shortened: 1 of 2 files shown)


server/mergin/tests/test_private_project_api.py
```diff
@@ -423,37 +425,76 @@ def test_admin_project_list(client):
...
+                version=version if version else "",
+            )
         )
-    )
-    if expected == 200:
-        os.remove(project_version.zip_path)
-    assert resp.status_code == expected
-    assert mock_create_zip.called if expected == 202 else not mock_create_zip.called
-    if not version and expected != 200:
+    finally:
```
varmar05: I think it is better to do test cleanups at the beginning
- you have results from failed tests
- you are sure to have a clean state for new test

PR #498, opened 2025-08-15
Cleanup of non necessary entrypoints

MarcelGeo:
- increase heaer size for gunicorn
- cleanup duplicated MERGIN_BASE_URL - it's hidden in middle of a file and customers then don't know why they have fixme in invitations emails

### DIFF
--- a/deployment/enterprise/.env.template
+++ b/deployment/enterprise/.env.template
@@ -176,7 +176,7 @@ ACCOUNT_EXPIRATION=1
 # for links generated in emails
 
 #MERGIN_LOGO_URL=  # for link to logo in emails
-MERGIN_LOGO_URL=fixme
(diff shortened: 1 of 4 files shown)


deployment/enterprise/.env.template
```diff
@@ -175,9 +175,6 @@ ACCOUNT_EXPIRATION=1
 
 # for links generated in emails
 
-#MERGIN_BASE_URL=http://localhost:5000
-MERGIN_BASE_URL=fixme
```
varmar05, line 179: hm, we need to make sure this make sense and is always set - maybe we should remove default value then

varmar05 (changes requested):

deployment/enterprise/.env.template
MarcelGeo: this is duplicated env variable in .env file

fernandinand:
Looks good!

PR #499, opened 2025-08-18
Add post request to explicitely prepare download archive

harminius:
Follows up https://github.com/MerginMaps/server/pull/497

Hitting the download button:
1.  `POST` - triggers the celery job to create the project zip
2. `HEAD` - polls server until the archive is ready (`200`) and 
3. hidden `GET` downloads the archive

:heavy_plus_sign: Improvements:
Only one celery job can be created from a browser

### DIFF
--- a/server/mergin/sync/private_api_controller.py
+++ b/server/mergin/sync/private_api_controller.py
@@ -331,9 +331,6 @@ def download_project(id: str, version=None):  # noqa: E501 # pylint: disable=W06
         project_id=project.id, name=lookup_version
     ).first_or_404("Project version does not exist")
 
-    if project_version.project_size > current_app.config["MAX_DOWNLOAD_ARCHIVE_SIZE"]:
-        abort(400)
-
     # check zip is already created
     if os.path.exists(project_version.zip_path):
         if current_app.config["USE_X_ACCEL"]:
@@ -370,7 +367,7 @@ def prepare_archive(id: str, version=None):
         abort(400)
 
     if os.path.exists(pv.zip_path):
-        return "Project archive already ready", 200
+        return NoContent, 204
 
     # trigger job if no recent partial
     temp_zip_path = pv.zip_path + ".partial"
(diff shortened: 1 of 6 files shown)


server/mergin/sync/private_api_controller.py
```diff
@@ -352,17 +352,35 @@ def download_project(id: str, version=None):  # noqa: E501 # pylint: disable=W06
...
+    """Triggers background job to create project archive"""
+    project = require_project_by_uuid(id, ProjectPermissions.Read)
+    lookup_version = (
+        ProjectVersion.from_v_name(version) if version else project.latest_version
+    )
+    pv = ProjectVersion.query.filter_by(
+        project_id=project.id, name=lookup_version
+    ).first_or_404()
+
+    if pv.project_size > current_app.config["MAX_DOWNLOAD_ARCHIVE_SIZE"]:
```
varmar05, line 366: I think we can remove such condition from GET endpoint then

server/mergin/sync/private_api_controller.py
```diff
@@ -352,17 +352,35 @@ def download_project(id: str, version=None):  # noqa: E501 # pylint: disable=W06
...
+        abort(400)
+
+    if os.path.exists(pv.zip_path):
+        return "Project archive already ready", 200
```
varmar05: hm, depends on the point of view, if the only expected action on this endpoint it to _trigger_ the job, I would return 204 and no content or even 409 as there is already required resource on server

web-app/packages/lib/src/modules/project/store.ts
```diff
@@ -707,6 +707,7 @@ export const useProjectStore = defineStore('projectModule', {
     },
 
     async downloadArchive(payload: DownloadPayload) {
+      console.log("You are here")
```
varmar05: remove

server/mergin/sync/private_api_controller.py
```diff
@@ -352,17 +352,35 @@ def download_project(id: str, version=None):  # noqa: E501 # pylint: disable=W06
...
+    partial_exists = os.path.exists(temp_zip_path)
+    is_expired = partial_exists and datetime.fromtimestamp(
+        os.path.getmtime(temp_zip_path), tz=timezone.utc
+    ) < datetime.now(timezone.utc) - timedelta(
+        seconds=current_app.config["PARTIAL_ZIP_EXPIRATION"]
+    )
+    if not partial_exists or is_expired:
+        create_project_version_zip.delay(pv.id)
+
+    return "Project zip being prepared", 202
```
varmar05: I don't think we need 'pending' status here as this job simply creates a request, so I would use 200/201 and leave next GET/HEAD to do polling.

RastoHu:
test: download project from web

PR #504, opened 2025-09-01
Merged diffs

varmar05:


server/mergin/sync/utils.py
```diff
@@ -577,3 +580,105 @@ def get_x_accel_uri(*url_parts):
...
+
+    mime_type = "text/plain" if not is_binary(abs_path) else get_mimetype(abs_path)
+    resp.headers["Content-Type"] = mime_type
+    file_name = quote(os.path.basename(path).encode("utf-8"))
+    resp.headers["Content-Disposition"] = f"attachment; filename*=UTF-8''{file_name}"
+    resp.direct_passthrough = False
+    return resp
+
+
+LOG_BASE = 4
```
MarcelGeo: Just :nail_care: - add on top with some comment

server/mergin/sync/utils.py
```diff
@@ -577,3 +580,105 @@ def get_x_accel_uri(*url_parts):
...
+        if version % LOG_BASE**rank:
+            continue
+
+        index = version // LOG_BASE**rank
+        levels.append(CachedLevel(rank=rank, index=index))
+
+    return levels
+
+
+def get_merged_versions(start: int, end: int) -> List[CachedLevel]:
```
MarcelGeo: Trying to check this:
Is it ok to not return partial levels , like 1-16 etc. ? 
```
print(get_merged_versions(1,64))
>>> [CachedLevel(rank=3, index=1, versions=v1-v64)]
```

server/mergin/sync/utils.py
```diff
@@ -577,3 +580,105 @@ def get_x_accel_uri(*url_parts):
...
+    Get all (merged) versions between start version and end version while respecting cached levels.
+    This basically provide the list of smaller versions (checkpoints) to be merged in order to get the final version.
+    """
+    levels = []
+    while start <= end:
+        if start == end:
+            rank_max = 0
+        else:
+            rank_max = math.floor((math.log(end - start + 1) / math.log(LOG_BASE)))
+        for rank in reversed(range(0, rank_max + 1)):
```
MarcelGeo: we could use also `range(rank_max + 1, -1, -1)`, but just :nail_care:

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -128,3 +129,20 @@ def remove_project_collaborator(id, user_id):
...
+
+def download_diff_file(id: str, file: str):
+    """Download project geopackage diff file"""
+    project = require_project_by_uuid(id, ProjectPermissions.Read)
+    diff_file = FileDiff.query.filter_by(path=file).first_or_404()
+
+    # create merged diff if it does not exist
+    if not os.path.exists(diff_file.abs_path):
+        diff_file.construct_checkpoint()
+        if not os.path.exists(diff_file.abs_path):
```
MarcelGeo: Is this safe check necessary?

server/mergin/sync/utils.py
```diff
@@ -7,13 +7,16 @@
 import hashlib
 import re
 import secrets
+from binaryornot.check import is_binary
```
MarcelGeo, line 12: Do we have deps for that?

server/mergin/sync/utils.py
```diff
@@ -577,3 +580,105 @@ def get_x_accel_uri(*url_parts):
...
+        return str(self)
+
+
+def get_cached_levels(version: int) -> List[CachedLevel]:
+    """
+    Return the most right part of version tree as other nodes are already cached.
+    Version must divisible by BASE, and then we calculate all cached levels related to it.
+    """
+    levels = []
+    rank_max = math.floor((math.log(version) / math.log(LOG_BASE)))
```
MarcelGeo: Can we simplify this to `math.log(version, LOG_BASE)` ?

server/mergin/tests/test_file_restore.py
```diff
@@ -183,25 +178,8 @@ def test_version_file_restore(diff_project):
     diff_project.storage.restore_versioned_file("base.gpkg", 6)
     assert os.path.exists(test_file)
     assert gpkgs_are_equal(test_file, test_file + "_backup")
-
-    # remove v9 and v10 to mimic that project history end with existing file
-    pv_8 = ProjectVersion.query.filter_by(project_id=diff_project.id, name=8).first()
```
MarcelGeo, line 188: Interersting, that removed.

server/mergin/sync/tasks.py
MarcelGeo, line 1: Cleanup deps import

server/migrations/community/bd1ec73db389_create_file_diff_table.py
```diff
@@ -27,8 +27,8 @@ def upgrade():
         sa.Column("rank", sa.Integer(), nullable=False),
         sa.Column("version", sa.Integer(), nullable=False),
         sa.Column("location", sa.String(), nullable=True),
-        sa.Column("size", sa.BigInteger(), nullable=False),
-        sa.Column("checksum", sa.String(), nullable=False),
+        sa.Column("size", sa.BigInteger(), nullable=True),
```
MarcelGeo, line 30: Let's document that properly for clients , if not... that size will be not available in some time.

server/mergin/sync/models.py
```diff
@@ -734,22 +735,24 @@ class FileDiff(db.Model):
...
-    checksum = db.Column(db.String, nullable=False)
+    size = db.Column(db.BigInteger, nullable=True)
+    checksum = db.Column(db.String, nullable=True)
 
     __table_args__ = (
         db.UniqueConstraint("file_path_id", "rank", "version", name="unique_diff"),
         db.Index("ix_file_diff_file_path_id_version_rank", file_path_id, version, rank),
     )
 
+    file = db.relationship("ProjectFilePath", uselist=False)
```
MarcelGeo, line 752: This is not related to migration?

server/mergin/sync/models.py
```diff
@@ -758,11 +761,140 @@ def __init__(
...
+        if self.rank == 0:
+            return CachedLevel(rank=self.rank, index=self.version)
+
+        # merged diffs can only be created for certain versions
+        if self.version % LOG_BASE:
+            return
+
+        index = self.version // LOG_BASE**self.rank
+        # some invalid record
+        if index < 1 or self.rank < 0:
```
MarcelGeo: Use try catch instead for value error in CachedLevel.

server/mergin/sync/models.py
```diff
@@ -758,11 +761,140 @@ def __init__(
...
+
+    def construct_checkpoint(self) -> None:
+        """Create a diff file checkpoint (aka. merged diff).
+        Find all smaller diffs which are needed to create the final diff file and merge them.
+        In case of missing some lower rank checkpoint, use individual diffs instead.
+        """
+        if os.path.exists(self.abs_path):
+            return
+
+        basefile = FileHistory.get_basefile(self.file_path_id, self.cache_level.end)
```
MarcelGeo: Do we need such cache_level property? It looks like one timer directly in this construct_checkpoint

server/mergin/sync/public_api_v2_controller.py
MarcelGeo: I think If there is not check point or unsuccessful checkpoint , we do not want to make 404 ... hm?

server/mergin/sync/utils.py
varmar05: The method should give you the list of `CachedLevel` you would need to apply to `start` to end up with gpkg at `end` version. So in your case it would be just single large diff because you exactly matched the range in caching tree.

Maybe we should rename function and update docstring to make it clearer. Open to suggestions.

server/mergin/tests/test_file_restore.py
varmar05: this test was about trying 'backward search'  to reconstruct gpkg file - such use case does not exist anymore

server/mergin/sync/utils.py
varmar05: this was just moved from other place to utils - we are already using that

server/mergin/sync/public_api_v2_controller.py
varmar05: yea, we can return anything else - there must be something broken in diffs if we cannot reconstruct it

what do you suggest, some logging.error and 422? At the end download of this diff would probably never succeed.

server/mergin/sync/utils.py
varmar05: not sure if that would be more readable though

server/mergin/sync/models.py
varmar05: it is just convenience ORM mapping

server/mergin/sync/models.py
varmar05: I thought it would be useful to have it but we can get rid of it

server/mergin/sync/public_api_v2_controller.py
MarcelGeo: yes, or just 400 :)

varmar05: changed this to 422 and custom error so client can implement some error handling there

MarcelGeo (approved):

PR #506, opened 2025-09-03
Return whole project nfo dump from project versions

MarcelGeo:
TO ensure that project info dump will be properly saved in clients .mergin folders, we are returning current project info created from version instead of project version.

### DIFF
--- a/server/mergin/sync/public_api_v2_controller.py
+++ b/server/mergin/sync/public_api_v2_controller.py
@@ -39,7 +39,12 @@ from .models import (
 )
 from .permissions import ProjectPermissions, require_project_by_uuid
 from .public_api_controller import catch_sync_failure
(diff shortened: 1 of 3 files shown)


server/mergin/sync/public_api_v2_controller.py
```diff
@@ -337,17 +342,7 @@ def create_project_version(id):
...
-        ProjectVersionSchema(
-            exclude=(
-                "files",
-                "changes",
-                "changesets",
-            )
-        ).dump(pv),
-        201,
-    )
+    return ProjectSchema().dump(project), 201
```
varmar05, line 345: we discussed this with @tomasMizera and it seems not all data is needed, e.g. history in files field.
Could you double check we return only what's needed? (Probably inline with new v2 project info)

MarcelGeo: ok we can do it in separated PR / ticket.

PR #514, opened 2025-09-24
Release EE/CE 2025.7.3 config

MarcelGeo:


deployment/enterprise/docker-compose.maps.yml
```diff
@@ -30,13 +30,15 @@ services:
...
-    image: 433835555346.dkr.ecr.eu-west-1.amazonaws.com/mergin/qgis-extractor-ee:2025.1.0
+    image: 433835555346.dkr.ecr.eu-west-1.amazonaws.com/mergin/qgis-extractor-ee:2025.3.0
     user: 901:999
     networks:
       - mergin-net
     environment:
       - OVERVIEWS_DATA_DIR=/data
       - MM_WMS_TILE_BUFFER=100
       - MM_WMS_AVOID_ARTIFACTS=1
+      - BROKER_URL=redis://mergin-redis:6379/0
```
varmar05: Don't we need to remove some configs from previous extractor / maps celery?

fernandinand: The container name is not consistent with main compose file. I would rather use service name here (`redis`)

deployment/enterprise/docker-compose.maps.yml
```diff
@@ -30,13 +30,15 @@ services:
...
+      - CELERY_RESULT_BACKEND=redis://mergin-redis:6379/0
```
fernandinand: The container name is not consistent with main compose file. I would rather use service name here (`redis`)

fernandinand (changes requested):
Have service name on `BROKER_URL` env on qgis_extractor service

PR #520, opened 2025-10-09
Checkpoints introduction for version changes

MarcelGeo:
https://github.com/MerginMaps/server-private/issues/3014

We are introducing new table for storing chckpoints (project_version_change), where rank and delta json for current version is stored. If version are in logarithms logic with base=4, then store higher level checkpoints if needed.

Structure for delta json: List[DeltaData]. If there is not diff, diff key is not provided in result.

```json
[
  {
    "diff": "Survey_points.gpkg-diff-14ee7fe7-366a-4ba8-85f3-f15c38242d93",
    "path": "Survey_points.gpkg",
    "size": 98304,
    "change": "update_diff",
    "version":  25,
    "checksum": "12b26ef4ed11e4fb411baf40402fedf26696f830"
  }
]
```

**Delta creation logic**
- if new version is created, Project version change with delta and rank=0 is created
- if somebody will return changes which will be aplied in /delta endpoint, checkpoints are created or returned if exists
- as side effect, If checkpoint does not exists, FileDiff new file future geodiff file name is generated in FileDiff table

**Migration**

Migrate file history and file diffs to merged delta in project_version_change table.

### DIFF
--- a/server/mergin/sync/files.py
+++ b/server/mergin/sync/files.py
@@ -331,8 +331,8 @@ class DeltaChangeSchema(DeltaChangeBaseSchema):
         return data
 
 
-class DeltaChangeRespSchema(DeltaChangeBaseSchema):
-    """Schema for delta data response"""
+class DeltaChangeItemSchema(DeltaChangeBaseSchema):
+    """Schema for delta changes response"""
 
     diffs = fields.List(fields.Nested(DeltaChangeDiffFileSchema()))
 
@@ -342,3 +342,12 @@ class DeltaChangeRespSchema(DeltaChangeBaseSchema):
         if not data.get("diffs"):
             data.pop("diffs", None)
         return data
+
+
+class DeltaChangeRespSchema(ma.Schema):
+    """Schema for list of delta changes wrapped in items field"""
+
+    items = fields.List(fields.Nested(DeltaChangeItemSchema()))
+
+    class Meta:
+        unknown = EXCLUDE
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -65,6 +65,16 @@ class FileSyncErrorType(Enum):
     SYNC_ERROR = "sync error"
 
 
+class ChangeComparisonAction(Enum):
+    """Actions to take when comparing two changes"""
+
+    REPLACE = "replace"
+    DELETE = "delete"
+    UPDATE = "update"
+    UPDATE_DIFF = "update_diff"
+    EXCLUDE = "exclude"  # Return None to exclude the file
+
+
 class Project(db.Model):
     id = db.Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     name = db.Column(db.String, index=True)
@@ -993,7 +1003,7 @@ class ProjectVersionDelta(db.Model):
     version = db.Column(db.Integer, nullable=False, index=True)
     # exponential order of changes json
     rank = db.Column(db.Integer, nullable=False, index=True)
-    # to which project version is this linked
+    # to which project is this linked
     project_id = db.Column(
         UUID(as_uuid=True),
         db.ForeignKey("project.id", ondelete="CASCADE"),
@@ -1025,94 +1035,167 @@ class ProjectVersionDelta(db.Model):
         Merge changes json array objects into one list of changes.
         Changes are merged based on file path and change type.
         """
-        result: Dict[str, DeltaChangeMerged] = {}
         updating_files: Set[str] = set()
         # sorting changes by version to apply them in correct order
         items.sort(key=lambda x: x.version)
 
-        def handle_replace(result, path, current, previous):
-            result[path] = current
-
-        def handle_delete(result, path, current, previous):
-
-            if path in updating_files:
-                result[path] = current
-            else:
-                del result[path]
-
-        def handle_update(result, path, current, previous):
-            # handle update case, when previous change was create - just revert to create with new metadata
-            current.change = previous.change
-            current.diffs = []
-            result[path] = current
-
-        def handle_update_diff(result, path, current, previous):
-            current.diffs = (previous.diffs or []) + (current.diffs or [])
-            result[path] = current
-
-        dispatch = {
-            # create + delete = file is transparent for current changes -> delete it
-            (PushChangeType.CREATE, PushChangeType.DELETE): handle_delete,
-            # create + update = create with updated info
-            (PushChangeType.CREATE, PushChangeType.UPDATE): handle_update,
-            (PushChangeType.CREATE, PushChangeType.UPDATE_DIFF): handle_update,
-            (PushChangeType.CREATE, PushChangeType.CREATE): None,
-            # update + update_diff = update with latest info
-            (
-                PushChangeType.UPDATE,
-                PushChangeType.UPDATE_DIFF,
-            ): handle_update,
-            (PushChangeType.UPDATE, PushChangeType.UPDATE): handle_replace,
-            (PushChangeType.UPDATE, PushChangeType.DELETE): handle_replace,
-            (PushChangeType.UPDATE, PushChangeType.CREATE): handle_replace,
-            # update_diff + update_diff = update_diff with latest info with proper order of diffs
-            (
-                PushChangeType.UPDATE_DIFF,
-                PushChangeType.UPDATE_DIFF,
-            ): handle_update_diff,
-            (PushChangeType.UPDATE_DIFF, PushChangeType.UPDATE): handle_replace,
-            (PushChangeType.UPDATE_DIFF, PushChangeType.DELETE): handle_replace,
-            (PushChangeType.UPDATE_DIFF, PushChangeType.CREATE): None,
-            (PushChangeType.DELETE, PushChangeType.CREATE): handle_replace,
-            # delete + update = invalid sequence, keep delete
-            (PushChangeType.DELETE, PushChangeType.UPDATE): None,
-            (PushChangeType.DELETE, PushChangeType.UPDATE_DIFF): None,
-            (PushChangeType.DELETE, PushChangeType.DELETE): None,
-        }
-
+        # Merge changes for each file in a single pass
+        result: Dict[str, DeltaChangeMerged] = {}
         for item in items:
+            path = item.path
             current = item.to_merged()
-            path = current.path
-            # path is key for merging changes
-            previous = result.get(path)
 
-            # adding new file change if not seen before
-            if not previous:
+            # First change for this file
+            if path not in result:
                 result[path] = current
-                # track existing paths to avoid deleting created files later
+                # track existing paths to avoid deleting files that are already in history before
                 if current.change != PushChangeType.CREATE:
                     updating_files.add(path)
                 continue
 
-            handler = dispatch.get((previous.change, current.change))
-            if handler:
-                handler(result, path, current, previous)
+            # Compare and merge with previous change for this file
+            can_delete = path in updating_files
+            new_change = ProjectVersionDelta._compare_changes(
+                result[path], current, can_delete
+            )
+
+            # Update result (or remove if no change is detected)
+            if new_change is not None:
+                result[path] = new_change
+            else:
+                del result[path]
 
         return list(result.values())
+
+    @staticmethod
+    def _compare_changes(
+        previous: DeltaChangeMerged,
+        new: DeltaChangeMerged,
+        prevent_delete_change: bool,
+    ) -> Optional[DeltaChangeMerged]:
+        """
+        Compare and merge two changes for the same file.
+
+        Args:
+            previous: Previously accumulated change
+            new: New change to compare
+            prevent_delete_change: Whether the change can be deleted when resolving create+delete sequences
+
+        Returns:
+            Merged change or None if file should be excluded
+        """
+
+        # Map change type pairs to actions
+        action_map = {
+            # create + delete = file is transparent for current changes -> delete it
+            (
+                PushChangeType.CREATE,
+                PushChangeType.DELETE,
+            ): ChangeComparisonAction.DELETE,
+            # create + update = create with updated info
+            (
+                PushChangeType.CREATE,
+                PushChangeType.UPDATE,
+            ): ChangeComparisonAction.UPDATE,
+            (
+                PushChangeType.CREATE,
+                PushChangeType.UPDATE_DIFF,
+            ): ChangeComparisonAction.UPDATE,
+            (
+                PushChangeType.CREATE,
+                PushChangeType.CREATE,
+            ): ChangeComparisonAction.EXCLUDE,
+            # update + update_diff = update with latest info
+            (
+                PushChangeType.UPDATE,
+                PushChangeType.UPDATE_DIFF,
+            ): ChangeComparisonAction.UPDATE,
+            (
+                PushChangeType.UPDATE,
+                PushChangeType.UPDATE,
+            ): ChangeComparisonAction.REPLACE,
+            (
+                PushChangeType.UPDATE,
+                PushChangeType.DELETE,
+            ): ChangeComparisonAction.REPLACE,
+            (
+                PushChangeType.UPDATE,
+                PushChangeType.CREATE,
+            ): ChangeComparisonAction.REPLACE,
+            # update_diff + update_diff = update_diff with latest info with proper order of diffs
+            (
+                PushChangeType.UPDATE_DIFF,
+                PushChangeType.UPDATE_DIFF,
+            ): ChangeComparisonAction.UPDATE_DIFF,
+            (
+                PushChangeType.UPDATE_DIFF,
+                PushChangeType.UPDATE,
+            ): ChangeComparisonAction.REPLACE,
+            (
+                PushChangeType.UPDATE_DIFF,
+                PushChangeType.DELETE,
+            ): ChangeComparisonAction.REPLACE,
+            (
+                PushChangeType.UPDATE_DIFF,
+                PushChangeType.CREATE,
+            ): ChangeComparisonAction.EXCLUDE,
+            (
+                PushChangeType.DELETE,
+                PushChangeType.CREATE,
+            ): ChangeComparisonAction.REPLACE,
+            # delete + update = invalid sequence
+            (
+                PushChangeType.DELETE,
+                PushChangeType.UPDATE,
+            ): ChangeComparisonAction.EXCLUDE,
+            (
+                PushChangeType.DELETE,
+                PushChangeType.UPDATE_DIFF,
+            ): ChangeComparisonAction.EXCLUDE,
+            (
+                PushChangeType.DELETE,
+                PushChangeType.DELETE,
+            ): ChangeComparisonAction.EXCLUDE,
+        }
+
+        action = action_map.get((previous.change, new.change))
+        result = None
+        if action == ChangeComparisonAction.REPLACE:
+            result = new
+
+        elif action == ChangeComparisonAction.DELETE:
+            # if change is create + delete, we can just remove the change from accumulated changes
+            # only if this action is allowed (file existed before)
+            if prevent_delete_change:
+                result = new
+
+        elif action == ChangeComparisonAction.UPDATE:
+            # handle update case, when previous change was create - just revert to create with new metadata
+            new.change = previous.change
+            new.diffs = []
+            result = new
+
+        elif action == ChangeComparisonAction.UPDATE_DIFF:
+            new.diffs = (previous.diffs or []) + (new.diffs or [])
+            result = new
+
+        return result
 
     @classmethod
     def create_checkpoint(
         cls,
         project_id: str,
         checkpoint: Checkpoint,
-        changes: List[ProjectVersionDelta] = [],
+        from_deltas: List[ProjectVersionDelta] = [],
     ) -> Optional[ProjectVersionDelta]:
         """
         Creates and caches new checkpoint and any required FileDiff checkpoints.
+        Use from_deltas to create checkpoint from existing individual deltas.
+        Returns created ProjectVersionDelta object with checkpoint.
         """
         delta_range = [
             change
-            for change in changes
+            for change in from_deltas
             if checkpoint.start <= change.version <= checkpoint.end
         ]
 
@@ -1122,7 +1205,7 @@ class ProjectVersionDelta(db.Model):
             )
             return None
 
-        # dump delta objects from database and flatten list for merging
+        # dump changes lists from database and flatten list for merging
         changes = []
         for delta in delta_range:
             changes.extend(DeltaChangeSchema(many=True).load(delta.changes))
--- a/server/mergin/sync/public_api_v2.yaml
+++ b/server/mergin/sync/public_api_v2.yaml
@@ -410,9 +410,7 @@ paths:
           content:
             application/json:
               schema:
-                type: array
-                items:
-                  $ref: "#/components/schemas/ProjectDelta"
+                $ref: "#/components/schemas/ProjectDeltaResponse"
         "400":
           $ref: "#/components/responses/BadRequest"
         "404":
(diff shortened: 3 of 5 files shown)


server/mergin/sync/files.py
```diff
@@ -234,3 +241,98 @@ def patch_field(self, data, **kwargs):
...
+            change=self.change,
+            version=self.version,
+        )
+        if self.diffs:
+            result.diff = self.diffs[0].path
+        return result
+
+
+@dataclass
+class DeltaData(File):
```
varmar05: This can inherit `DeltaBase`

server/mergin/sync/files.py
```diff
@@ -234,3 +241,98 @@ def patch_field(self, data, **kwargs):
...
+class DeltaDiffFile:
+    path: str
+
+
+class DeltaDiffFileSchema(ma.Schema):
+    path = fields.String(required=True)
+
+
+@dataclass
+class DeltaBase(File):
```
varmar05: It would be good to comment somewhere what this 'Delta' actually means (for future reference).

server/mergin/sync/files.py
```diff
@@ -234,3 +241,98 @@ def patch_field(self, data, **kwargs):
...
+            checksum=self.checksum,
+            change=self.change,
+            version=self.version,
+        )
+        if self.diff:
+            result.diffs = [DeltaDiffFile(path=self.diff)]
+        return result
+
+
+class DeltaBaseSchema(ma.Schema):
```
varmar05: maybe this could be in schemas next to other api schemas?

server/mergin/sync/files.py
```diff
@@ -234,3 +241,98 @@ def patch_field(self, data, **kwargs):
...
+    """Base schema for detla json and response from delta endpoint"""
```
varmar05: 'detla' - typo

server/mergin/sync/files.py
```diff
@@ -234,3 +241,98 @@ def patch_field(self, data, **kwargs):
...
+
+    path = fields.String(required=True)
+    size = fields.Integer(required=True)
+    checksum = fields.String(required=True)
+    version = fields.Integer(required=True)
+    change = fields.Enum(PushChangeType, by_value=True, required=True)
+
+
+class DeltaDataSchema(DeltaBaseSchema):
```
varmar05: do we even need this one? load/patch could go to ORM class

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
         return True
 
 
+class ProjectVersionChange(db.Model):
```
varmar05: how about ProjectDelta? 🤔

server/mergin/sync/models.py
```diff
@@ -997,6 +1243,28 @@ def __init__(
             else:
                 latest_files_map[fh.path] = fh.id
 
+        # cache changes data json for version checkpoints
+        # rank 0 is for all changes from start to current version
+        delta_data = [
+            DeltaData(
+                path=c.path,
```
varmar05, line 1371: hm, this pretty much looks like ProjectFileChange (apart of version)

server/mergin/sync/public_api_v2.yaml
```diff
@@ -381,6 +381,42 @@ paths:
...
+      parameters:
+        - $ref: "#/components/parameters/ProjectId"
+        - name: since
+          in: query
+          required: true
+          schema:
+            type: integer
+            example: 1
+          description: Start version (exclusive)
+        - name: to
```
varmar05, line 400: let's double check specs if it is not always latest version

server/mergin/sync/public_api_v2.yaml
```diff
@@ -800,3 +836,37 @@ components:
             - editor
             - writer
             - owner
+    ProjectDelta:
```
varmar05: This should be probably split to File (as it would come up from other endpoints) and Delta

server/mergin/sync/public_api_v2.yaml
```diff
@@ -800,3 +836,37 @@ components:
...
+          example: 1024
+        checksum:
+          type: string
+          example: 9adb76bf81a34880209040ffe5ee262a090b62ab
+        version:
+          type: integer
+          example: 1
+        change:
+          type: string
+          enum: [create, update, delete, update_diff]
```
varmar05: maybe better to define custom type for this

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -402,3 +402,28 @@ def upload_chunk(id: str):
         UploadChunkSchema().dump({"id": chunk_id, "valid_until": valid_until}),
         200,
     )
+
+
+@auth_required
+def get_project_delta(id: str):
+    """Get project changes (delta) between two versions"""
+    since = int(request.args.get("since"))
```
varmar05: put those as function args

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -402,3 +402,28 @@ def upload_chunk(id: str):
         UploadChunkSchema().dump({"id": chunk_id, "valid_until": valid_until}),
         200,
     )
+
+
+@auth_required
```
varmar05: because this will be used in project pull this should work for public projects and therefore without auth required

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -402,3 +402,28 @@ def upload_chunk(id: str):
...
+def get_project_delta(id: str):
+    """Get project changes (delta) between two versions"""
+    since = int(request.args.get("since"))
+    to = int(request.args.get("to"))
+    project = require_project_by_uuid(id, ProjectPermissions.Read)
+    if since < 0 or to < 0:
```
varmar05: can't we pass this to connextion openapi?

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -402,3 +402,28 @@ def upload_chunk(id: str):
...
+        abort(400, "Invalid 'since' or 'to' version")
+    if since > to:
+        abort(400, "'since' version must be less than 'to' version")
+
+    ProjectVersion.query.filter(
```
varmar05: Is not enough to check that 'to' is <= project.latest_version?  

Alternatively we could do a check in single query (and `.count()`)

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -402,3 +402,28 @@ def upload_chunk(id: str):
...
+        ProjectVersion.project_id == project.id,
+        ProjectVersion.name == since,
+    ).first_or_404()
+    ProjectVersion.query.filter(
+        ProjectVersion.project_id == project.id,
+        ProjectVersion.name == to,
+    ).first_or_404()
+
+    deltas = ProjectVersionChange.get_delta(project.id, since, to)
```
varmar05: hm, might be even good to have this in Project class

server/migrations/community/63adc90fca0c_add_project_version_changes.py
```diff
@@ -0,0 +1,109 @@
...
+        SELECT
+            h.version_id,
+            0 AS rank,
+            jsonb_agg(
+                jsonb_strip_nulls(
+                    jsonb_build_object(
+                        'path', fp.path,
+                        'size', h.size,
+                        'change', h.change,
+                        'version', 'v' || h.project_version_name,
```
varmar05: should be int

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+        Get changes between two versions, merging them if needed.
+        - create FileDiff checkpoints if needed
+        - create ProjectVersionChange checkpoints if needed with delta json
+        """
+        if since > to:
+            logging.error(
+                f"Start version {since} is higher than end version {to} - broken history"
+            )
+            return
+        if since == to:
```
varmar05: I though that since is excluded and based on our recent discussion we should return changes that happened since  then, eg. 1 to 4 contains 2, 3, 4. Therefore I would expect that changes between the same version is None.

I would return single version if `to - since = 1`

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                    ProjectVersion.name == to,
+                    ProjectVersionChange.rank == 0,
+                )
+                .first()
+            )
+            return [DeltaMerged(**item) for item in change.delta] if change else None
+
+        expected_checkpoints = Checkpoint.get_checkpoints(since + 1, to)
+        expected_changes: List[ProjectVersionChange] = (
+            ProjectVersionChange.query.join(ProjectVersion)
```
varmar05: hm, given the usage (and general clarity) it might be better to move from `version_id` to `project_id` + `version_name` instead...

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                (checkpoint.rank, checkpoint.end)
+            )
+
+            # we have change in database, just return delta data from it
+            if expected_change:
+                deltas = DeltaDataSchema(many=True).load(expected_change.delta)
+                result.extend(deltas)
+                continue
+
+            if checkpoint.rank > 0:
```
varmar05: I guess this assumes that 0 ranks are always in db...

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                individual_changes = (
+                    cls.query_changes(project_id, since, to, 0)
```
varmar05: do we need a method for this?

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                    if not individual_changes
```
varmar05: in which case we would reuse those individual changes?

server/mergin/sync/models.py
varmar05: I would add some comment here that we need to construct missing higher rank and we do it from individual parts (for simplicity)

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                    else individual_changes
+                )
+                new_checkpoint = cls._create_checkpoint(
+                    project_id, checkpoint, individual_changes
+                )
+                if new_checkpoint:
```
varmar05: if new checkpoint is not created shall we return some error?

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+        individual_changes: List[ProjectVersionChange] = []
+
+        result: List[DeltaData] = []
+        for checkpoint in expected_checkpoints:
+            expected_change = existing_changes_map.get(
+                (checkpoint.rank, checkpoint.end)
+            )
+
+            # we have change in database, just return delta data from it
+            if expected_change:
```
varmar05: it is the same block as in L114, let's reverse the logic, create checkpoint if missing and at the end append delta

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+            rank,
+        ),
+    )
+    version = db.relationship(
+        "ProjectVersion",
+        uselist=False,
+    )
+
+    @staticmethod
+    def merge_delta_items(
```
varmar05: this might be cleaner to have only for single file path and resolve mapping to dict in different place,
currently it is confusing that the input list contains deltas for multiple files.

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+        items: List[DeltaData],
+    ) -> List[DeltaMerged]:
+        """
+        Merge multiple changes json array objects into one list of changes.
+        Changes are merged based on file path and change type.
+        """
+        result: Dict[str, DeltaMerged] = {}
+        for item in items:
+            current = item.to_merged_delta()
+            existing = result.get(current.path)
```
varmar05: maybe `previous`?

Also for merging it is crucial that items in list are ordered by version ascending - shall we make sure it is correct? I know we have it ordered from DB query but in case of tiny bug somewhere consequence here would be a disaster.

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+            path = current.path
+            if existing:
+                # merge changes data jsons
+                if existing.change == PushChangeType.CREATE:
+                    if current.change == PushChangeType.DELETE:
+                        # create + delete = nothing
```
varmar05: yes, for client this is a special case that changes cancelled themselves and such file never existed - let's extend a comment a bit

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                        del result[path]
+                    elif current.change in (
+                        PushChangeType.UPDATE,
+                        PushChangeType.UPDATE_DIFF,
+                    ):
+                        # create + update = create with updated info
```
varmar05: yes, for client this is completely new file so any subsequent updates will appear as a new file to client, just with the most recent metadata - pleas update a comment

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                        current.change = existing.change
+                        current.diffs = []
+                    else:
+                        result[path] = current
```
varmar05: ideally this should never happen?  - create + create

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                        current.change = existing.change
+                        current.diffs = []
+                    result[path] = current
+                elif existing.change == PushChangeType.UPDATE_DIFF:
+                    if current.change == PushChangeType.UPDATE_DIFF:
+                        # update_diff + update_diff = update_diff with latest info
+                        current.diffs.extend(existing.diffs or [])
+                    result[path] = current
+                else:
+                    # delete + anything = anything
```
varmar05: I don't think so - after delete we should only have a create action, otherwise we have weird inconsistent history - I'd be more specific here.

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                elif existing.change == PushChangeType.UPDATE:
+                    if current.change == PushChangeType.UPDATE_DIFF:
+                        # update + update_diff = update_diff with latest info
+                        current.change = existing.change
+                        current.diffs = []
+                    result[path] = current
+                elif existing.change == PushChangeType.UPDATE_DIFF:
+                    if current.change == PushChangeType.UPDATE_DIFF:
+                        # update_diff + update_diff = update_diff with latest info
+                        current.diffs.extend(existing.diffs or [])
```
varmar05: should not we extend previous.diffs with current diff(s)?

server/mergin/sync/models.py
```diff
@@ -902,6 +905,250 @@ def construct_checkpoint(self) -> bool:
...
+                        PushChangeType.UPDATE_DIFF,
+                    ):
+                        # create + update = create with updated info
+                        current.change = existing.change
+                        current.diffs = []
+                    else:
+                        result[path] = current
+                elif existing.change == PushChangeType.UPDATE:
+                    if current.change == PushChangeType.UPDATE_DIFF:
+                        # update + update_diff = update_diff with latest info
```
varmar05: this will end up as `UPDATE` with latest info because client does not have force-updated gpkg version and it is pointless  to pass diffs back (so they will be ignored all way long) - please update comment

server/mergin/sync/files.py
MarcelGeo: Yes

server/mergin/sync/models.py
MarcelGeo: Could be

server/mergin/sync/models.py
MarcelGeo: If you have multiple checkpoint to create. You do not need to fetch every time

server/mergin/sync/models.py
MarcelGeo: ?

server/mergin/sync/models.py
MarcelGeo: yeah that was my first version. I wanted to avoid composite keys

server/mergin/sync/models.py
MarcelGeo: renamed to previous

server/mergin/sync/models.py
MarcelGeo: yes order is important, I think it's fine... as we are returning ordered everything from db. My idea was to sort it somehow based on version ... but not sure.

server/mergin/sync/models.py
MarcelGeo: I renamed it finnaly to following -> ProjectVersionDelta -> contains changes column ... probably better than before, not sure here :D

server/mergin/sync/models.py
MarcelGeo: There are checks for individual changes

server/mergin/sync/public_api_v2.yaml
MarcelGeo: not sure what you mean

server/mergin/sync/models.py
MarcelGeo: there is diff but not string.

server/mergin/sync/public_api_v2.yaml
MarcelGeo: ?

server/mergin/sync/public_api_v2.yaml
varmar05: I have a feeling that in api we did not want to specify target version as it will always be a project latest version - best to confirm with Tomas

server/mergin/sync/public_api_v2.yaml
varmar05: I mean it contain generic File-like keys which will be used in v2/project/detail endpoints - but we can leave it for later

server/mergin/sync/models.py
varmar05: Yea, I know we discussed it but from the usage and what that table actually means (project.delta) we don't even need anything from ProjectVersion, only its name and project_id. Still, I think it would be cleaner if the new table had those fields even at expense of composite key.

server/mergin/sync/models.py
varmar05: I meant that fallback logic is written (correctly) for higher orders. I am just wondering what would happen if from some reason 0 rank file is not found - I guess we should return some error.

server/mergin/sync/models.py
varmar05: But isn't every checkpoint in list which we need to create composed from different individual diffs?

server/mergin/sync/models.py
varmar05: yea, that was my point, I know we return everything ordered from DB but I would probably still order it here in python just to play it safe, e.g. if someone changed ordering or similar kind of 'bug fix' not being aware of this

server/mergin/sync/models.py
varmar05: the comment was about making sure that order of diffs in list is correct

server/mergin/sync/public_api_v2_controller.py
MarcelGeo: we can get rid of this check completely

server/mergin/sync/public_api_v2.yaml
MarcelGeo: we added to as optional param

server/mergin/sync/models.py
MarcelGeo: let's upgrade diffs existing first, current last

MarcelGeo: done

server/mergin/sync/models.py
MarcelGeo: resolved to composite

server/mergin/sync/models.py
```diff
@@ -902,6 +988,205 @@ def construct_checkpoint(self) -> bool:
...
+
+    @staticmethod
+    def merge_changes(
+        items: List[DeltaChange],
+    ) -> List[DeltaChangeMerged]:
+        """
+        Merge changes json array objects into one list of changes.
+        Changes are merged based on file path and change type.
+        """
+        result: Dict[str, DeltaChangeMerged] = {}
```
varmar05: I think it would be more readable if we have this merging function defined on single file (path) - so we would pass just list of changes, order it and apply rules, then return final DeltaChange which would be assigned to that file in some upper level function. 

Hence no manipulation with dictionaries in local functions would be needed.

server/mergin/sync/models.py
```diff
@@ -902,6 +988,205 @@ def construct_checkpoint(self) -> bool:
...
+        updating_files: Set[str] = set()
+        # sorting changes by version to apply them in correct order
+        items.sort(key=lambda x: x.version)
+
+        def handle_replace(result, path, current, previous):
+            result[path] = current
+
+        def handle_delete(result, path, current, previous):
+
+            if path in updating_files:
```
varmar05: this is quite unclear what it is meant to be doing

server/mergin/sync/models.py
```diff
@@ -902,6 +988,205 @@ def construct_checkpoint(self) -> bool:
...
+            current.diffs = []
+            result[path] = current
+
+        def handle_update_diff(result, path, current, previous):
+            current.diffs = (previous.diffs or []) + (current.diffs or [])
+            result[path] = current
+
+        dispatch = {
+            # create + delete = file is transparent for current changes -> delete it
+            (PushChangeType.CREATE, PushChangeType.DELETE): handle_delete,
```
varmar05: I would probably just do some mapping: changes to action (can be enum) and then based on action we would do some handling for list items (see comment above).

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -402,3 +403,21 @@ def upload_chunk(id: str):
...
+def get_project_delta(id: str, since: int, to: Optional[int] = None):
+    """Get project changes (delta) between two versions"""
+
+    project: Project = require_project_by_uuid(id, ProjectPermissions.Read)
+    to = project.latest_version if to is None else to
+    if to > project.latest_version:
+        abort(400, "'to' version exceeds latest project version")
+
+    if since >= to:
+        abort(400, "'since' version must be less than or equal to 'to' version")
```
varmar05: I am bit confused, can we have since=to or not?

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -402,3 +403,21 @@ def upload_chunk(id: str):
...
+
+    delta_changes = project.get_delta_changes(since, to)
+    if delta_changes is None:
+        abort(404)
```
varmar05: if there were not changes (not sure if possible) I would return {} rather than 404

server/mergin/sync/models.py
MarcelGeo: merge_changes

- give out mapping table for actions

server/mergin/sync/models.py
```diff
@@ -1025,94 +1035,167 @@ def merge_changes(
...
-            result[path] = current
+        return list(result.values())
 
-        def handle_update_diff(result, path, current, previous):
-            current.diffs = (previous.diffs or []) + (current.diffs or [])
-            result[path] = current
+    @staticmethod
+    def _compare_changes(
+        previous: DeltaChangeMerged,
+        new: DeltaChangeMerged,
```
varmar05, line 1073: maybe `next` would be the better naming

PR #525, opened 2025-10-27
Disallow non-ASCII in email domain

harminius:
fixes https://github.com/MerginMaps/server-private/issues/3111

Non-ASCII characters in email domain might break email service provide API

We use close approximation of the regex provided by Brevo (https://github.com/MerginMaps/server-private/issues/3111#issuecomment-3491438139)

### DIFF
--- a/server/Pipfile
+++ b/server/Pipfile
@@ -41,7 +41,6 @@ importlib-metadata = "==8.4.0"  # https://github.com/pallets/flask/issues/4502
 typing_extensions = "==4.12.2"
 python-magic = "==0.4.27"
 click = "==8.2.0"
-regex = "==2025.11.3"
 # requirements for development on windows
 colorama = "==0.4.5"
 


server/mergin/auth/forms.py
```diff
@@ -48,18 +49,28 @@ class ExtendedEmail(Email):
...
-    Custom check for additional invalid characters disallows |'— because they make our email sending service to fail
+    4, leading, trailing, or consecutive dots in the local part,
+    5, invalid domain part - missing top level domain (user@example), consecutive dots,
+    Custom check for
+    - additional invalid characters disallows |'—
+    - non-ASCII characters in the domain part
+    because they make our email sending service to fail
     """
 
+    EMAIL_PATTERN = regex.compile(
```
varmar05: hm, are we certain we must switch to 3rd party lib, is there no workaround for what we need?

server/mergin/tests/test_auth.py
```diff
@@ -936,12 +937,12 @@ def test_server_usage(client):
     ("日人日本人", True),  # non-ascii character
     ("usér", True),  # non-ascii character
     ("user\\", False),  # disallowed character
-    ("user\260", True),  # non-ascii character (°)
+    ("user\260", False),  # not letter character (°)
```
varmar05, line 944: this is now forbidden?

server/mergin/auth/utils.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (C) Lutra Consulting Limited
+#
+# SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
+
+from email_validator import validate_email, EmailNotValidError
+
+
+def get_email_domain(email: str) -> str | None:
```
varmar05: is this still used?

server/mergin/tests/test_auth.py
harminius: yes, that was wrong - the thing is you can use non-ascii character in the local path but it must be a letter not a special character

server/mergin/auth/forms.py
harminius: found a close approximation with re that seems to work

server/mergin/auth/utils.py
harminius: rmed

PR #531, opened 2025-10-30
Project info v2

harminius:
Resolves https://github.com/MerginMaps/server-private/issues/3087

Create endpoint
- new marshmallow schema
- controller function

image

### DIFF
--- a/server/mergin/sync/public_api_v2_controller.py
+++ b/server/mergin/sync/public_api_v2_controller.py
@@ -14,7 +14,7 @@ from flask_login import current_user
 from marshmallow import ValidationError
 from sqlalchemy.exc import IntegrityError
 
-from .schemas_v2 import ProjectSchema
+from .schemas_v2 import ProjectSchema as ProjectSchemaV2
 from ..app import db
 from ..auth import auth_required
 from ..auth.models import User
@@ -166,7 +166,7 @@ def remove_project_collaborator(id, user_id):
 def get_project(id, files_at_version=None):
     """Get project info. Include list of files at specific version if requested."""
     project = require_project_by_uuid(id, ProjectPermissions.Read)
-    data = ProjectSchema().dump(project)
+    data = ProjectSchemaV2().dump(project)
     if files_at_version:
         pv = ProjectVersion.query.filter_by(
             project_id=project.id, name=ProjectVersion.from_v_name(files_at_version)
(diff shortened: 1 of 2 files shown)


server/mergin/sync/public_api_v2_controller.py
```diff
@@ -128,3 +130,18 @@ def remove_project_collaborator(id, user_id):
...
+
+
+def get_project(id, files_at_version=None):
+    """Get project info. Include list of files at specific version if requested."""
+    project = require_project_by_uuid(id, ProjectPermissions.Read)
+    data = ProjectSchema().dump(project)
+    if files_at_version:
+        pv = ProjectVersion.query.filter_by(
+            project_id=project.id, name=ProjectVersion.from_v_name(files_at_version)
+        ).first_or_404()
```
varmar05: hm, if some invalid version is passed, shall we return 404 or just ignore 'files' key?

server/mergin/sync/schemas_v2.py
```diff
@@ -0,0 +1,50 @@
...
+    updated_at = DateTimeWithZ(attribute="updated")
+
+    workspace = fields.Function(
+        lambda obj: {"id": obj.workspace.id, "name": obj.workspace.name}
+    )
+    role = fields.Method("_role")
+
+    def _role(self, obj):
+        role = ProjectPermissions.get_user_project_role(obj, current_user)
+        if not role:
```
varmar05: hm, can we get to the state that user has no role (permissions) to that project?

varmar05 (commented):
Looking good.

server/mergin/sync/schemas_v2.py
harminius: theoretically not.. 
maybe some race conditions when your role has changed during after the initial permission check and before this schema dumping
- no problem to remove if you prefer

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -139,9 +139,10 @@ def get_project(id, files_at_version=None):
     if files_at_version:
         pv = ProjectVersion.query.filter_by(
             project_id=project.id, name=ProjectVersion.from_v_name(files_at_version)
-        ).first_or_404()
-        data["files"] = ProjectFileSchema(
-            only=("path", "mtime", "size", "checksum"), many=True
-        ).dump(pv.files)
+        ).first()
+        if pv:
```
varmar05, line 174: can you add test for this?

server/mergin/sync/schemas_v2.py
varmar05: we can keep it, no problem

varmar05 (approved):

PR #534, opened 2025-11-14
Refuse trailing whitespace in path

harminius:
Resolves https://github.com/MerginMaps/qgis-plugin/issues/693#issuecomment-3522173686
Resolve https://github.com/MerginMaps/server-private/issues/3134

We disallowed to create a new project in folder with trailing space path in plugin(https://github.com/MerginMaps/qgis-plugin/issues/693)
This PR blocks to sync (push) any files with whitespace at the end of a folder in the path.
To avoid breaking the synchronization of existing projects, we should check only newly added files.

image

### DIFF
--- a/server/mergin/sync/public_api_controller.py
+++ b/server/mergin/sync/public_api_controller.py
@@ -972,7 +972,7 @@ def push_finish(transaction_id):
         if len(unsupported_files):
             abort(
                 400,
-                f"Unsupported file type detected: {unsupported_files[0]}. "
+                f"Unsupported file type detected: '{unsupported_files[0]}'. "
                 f"Please remove the file or try compressing it into a ZIP file before uploading.",
             )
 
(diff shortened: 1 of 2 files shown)


server/mergin/sync/utils.py
```diff
@@ -338,14 +339,19 @@ def files_size():
 def is_valid_path(filepath: str) -> bool:
     """Check filepath and filename for invalid characters, absolute path or path traversal"""
     return (
-        not len(re.split(r"\.[/\\]", filepath)) > 1  # ./ or .\
+        not re.search(r"\.[/\\]", filepath)  # ./ or .\
```
MarcelGeo, line 337: Is the same like before?

server/mergin/tests/test_utils.py
```diff
@@ -228,6 +228,10 @@ def test_is_valid_path(client, filepath, allow):
     assert is_valid_path(filepath) == allow
 
 
+def test_has_trailing_space():
```
MarcelGeo: add more tests also success (false)

server/mergin/sync/utils.py
harminius: yes, it just seemed more readable to me

server/mergin/tests/test_utils.py
harminius: :ballot_box_with_check:

PR #535, opened 2025-11-14
Create checkpoint recursively

varmar05:
Rewrite fallback logic for create checkpoints functions (both merged diffs and project deltas) to create intermediate (lower ranks) checkpoints if missing recursively instead of using individual diffs directly.

Advantage of this approach is that in case of too heavy request which would not complete from individual diffs now it would at least cache some partial results.


MarcelGeo (approved):
Just wondering, If we do not want flask command where projectId will be as argument. This command will create checkpoints tree for some project. We do not need to manage it trough api call.

PR #537, opened 2025-11-24
Sync bottleneck: prevent remove chunks

MarcelGeo:


server/mergin/tests/test_public_api_v2.py
```diff
@@ -378,7 +378,7 @@ def test_create_version(client, data, expected, err_code):
     if expected == 201:
         assert response.json["version"] == "v2"
         assert project.latest_version == 2
-        assert all(not os.path.exists(chunk) for chunk in chunks)
+        assert all(os.path.exists(chunk) for chunk in chunks)
```
varmar05, line 389: Let's add some comment that chunks are not removed during push (after use) but we rely solely on clean up job. Also we might want to call that job here to confirm it works as expected.

MarcelGeo: yeah

server/mergin/tests/test_public_api_v2.py
```diff
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
 
+from mergin.sync.tasks import remove_unused_chunks
```
varmar05: I think the name of the function is a bit misleading now as we remove all chunks, e.g. `remove_outdated_chunks`

MarcelGeo: see commit bellow if ok.

server/mergin/sync/tasks.py
```diff
@@ -169,3 +169,12 @@ def remove_unused_chunks():
         if not os.path.isdir(dir):
             continue
         remove_outdated_files(dir, time_delta)
+
+
+@celery.task
+def remove_transaction_chunks(chunks=[]):
```
varmar05: although ok here I would not use mutable default arg as it is usually a bad practice, use typing + None (if we need default)

varmar05 (changes requested):

PR #541, opened 2025-11-28
Don't reveal private project exists to anonymous user

harminius:
fixes https://github.com/MerginMaps/server-private/issues/3087#issuecomment-3586086194

return `404` to anonymous user requesting private project info
a new flag ensures it is applied only in specific endpoints -> does not affect other responses but v2 get project info

add validation for version in API spec


server/mergin/sync/permissions.py
```diff
@@ -219,6 +219,9 @@ def require_project_by_uuid(uuid: str, permission: ProjectPermissions, scheduled
     if not scheduled:
         project = project.filter(Project.removed_at.is_(None))
     project = project.first_or_404()
+    # we don't want to tell anonymous user if a private project exists
+    if current_user.is_anonymous and not project.public:
```
varmar05: should not this be part of permissions.check method?

MarcelGeo: It will be ok to double check also `require_project` and other methods.

server/mergin/tests/test_public_api_v2.py
```diff
@@ -176,7 +176,13 @@ def test_get_project(client):
     test_workspace = create_workspace()
     project = create_project("new_project", test_workspace, admin)
     logout(client)
+    # anonymous user cannot access the private resource
+    response = client.get(f"v2/projects/{project.id}")
+    assert response.status_code == 404
     # lack of permissions
+    user = add_user("tests", "tests")
+    login(client, user.username, "tests")
+    time.sleep(1)
```
varmar05: why sleep?

server/mergin/tests/test_public_api_v2.py
```diff
@@ -176,7 +176,13 @@ def test_get_project(client):
...
     response = client.get(f"v2/projects/{project.id}")
     assert response.status_code == 403
```
varmar05, line 186: don't we want also 404 here?

server/mergin/sync/permissions.py
```diff
@@ -219,13 +233,18 @@ def require_project_by_uuid(uuid: str, permission: ProjectPermissions, scheduled
...
     project = project.first_or_404()
+
     workspace = project.workspace
     if not workspace:
         abort(404)
     if not is_active_workspace(workspace):
         abort(404, "Workspace doesn't exist")
     if not permission.check(project, current_user):
-        abort(403, "You do not have permissions for this project")
+        if expose:
```
varmar05: Should not this affect only Read perms check?

harminius: this version returned `404` also for authenticated users without permissions (the same as github https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#404-not-found-for-an-existing-resource)
We discussed on call that we will return `403` in this scenario..

server/mergin/sync/permissions.py
harminius: we don't want to return `403` there - `require_project` is not used in v2 endpoints

varmar05 (approved):

PR #543, opened 2025-11-28
Backend libs update

varmar05:
Upgrade Flask to 3.x and SqlAlchemy to 2.x alongside with related libs and dependencies.
Connexion is upgraded to 2.15 as 3.x introduces quite significant (and potentially breaking) changes.

The upgrade contains minimal changes to get the app running and still uses (legacy) db queries.


MarcelGeo (approved):
Just wondering:

- is it safe to change this MerginInfo class from db migrations point of view?
- is alembic compatible with this syntax by generating migrations, etc. ?

PR #546, opened 2025-12-11
Filter out specific filenames when cloning project

harminius:
Resolves https://github.com/MerginMaps/server-private/issues/3149
Parent https://github.com/MerginMaps/product/issues/104

A set of specific filenames is skipped when copying the files.

?
Exclude it from copying it to maps
add test

### DIFF
--- a/server/mergin/tests/test_project_controller.py
+++ b/server/mergin/tests/test_project_controller.py
@@ -2011,7 +2011,7 @@ versions_test_data = [
         {"page": 1, "per_page": 5, "desc": False},
         200,
         "v1",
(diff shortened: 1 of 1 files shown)


server/mergin/sync/public_api_controller.py
```diff
@@ -92,6 +92,10 @@
 from .errors import StorageLimitHit, ProjectLocked
 from ..utils import format_time_delta
 
+EXCLUDED_CLONE_FILENAMES = {
+    "qgis_cfg.xml",
```
varmar05: can we move this to sync config just in case we change our minds :-)

varmar05:
please add tests

PR #547, opened 2025-12-12
Added to_version as latest project version to delta reponse + another response updates

MarcelGeo:
When pulling transaction starts in clients, we need to know in which version s client operating to perform successfull project info and finish of pull. 

- to_version key added to response
- diff.path changed to diff.id in response as this path is abstract and for clients it's not path, just some file somewhere - it could be also name 

Other:

- improve handling of empty delta if since equals to to (or latest project version) -> returns items: [], bacuase clients will ping this endpoint everytime when somebody will try to pull.

### DIFF
--- a/server/mergin/sync/public_api_v2.yaml
+++ b/server/mergin/sync/public_api_v2.yaml
@@ -965,6 +965,7 @@ components:
       type: object
       required:
         - items
+        - to_version
       properties:
         to_version:
           $ref: "#/components/schemas/VersionName"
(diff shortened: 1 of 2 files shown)


server/mergin/sync/files.py
```diff
@@ -254,13 +254,13 @@ def patch_field(self, data, **kwargs):
...
 
-    path: str
+    id: str
 
 
 class DeltaChangeDiffFileSchema(ma.Schema):
     """Schema for diff file path in diffs list"""
 
-    path = fields.String(required=True)
+    id = fields.String(required=True)
```
varmar05, line 263: update also docstring

varmar05 (approved):

PR #551, opened 2026-01-09
Integrate last sign in

harminius:
Resolves https://github.com/MerginMaps/server-private/issues/3077

Updates
- collaborators tab - decrease number of columns and put all user name info together
- common `UserSummary` component created = avatar + fullname if present + username + email

Collaborators:
image

User detail in admin
Screenshot from 2026-01-12 16-52-17

### DIFF
--- a/web-app/packages/lib/src/common/components/UserSummary.vue
+++ b/web-app/packages/lib/src/common/components/UserSummary.vue
@@ -2,8 +2,8 @@
   <div class="flex align-items-center w-auto gap-2">
     <PAvatar
       :label="$filters.getAvatar(email, name)"
-      size="large"
       shape="circle"
+      :size="avatarSize"
       :pt="{
         root: {
           class: 'mr-1 text-color-forest font-semibold flex-shrink-0',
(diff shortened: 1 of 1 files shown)


server/mergin/sync/models.py
```diff
@@ -376,6 +378,7 @@ class ProjectAccessDetail:
     workspace_role: str
     project_role: Optional[ProjectRole]
     type: str
+    last_signed_in: str
```
varmar05: do we want this optional?

web-app/packages/lib/src/common/components/UserSummary.vue
```diff
@@ -0,0 +1,37 @@
...
+        root: {
+          class: 'mr-2 text-color-forest font-semibold flex-shrink-0',
+          style: { borderRadius: '50%' }
+        }
+      }"
+    />
+
+    <div class="flex flex-column justify-content-between">
+      <p class="title-t4 overflow-wrap-anywhere">
+        {{ fullName ? `${fullName} (${username})` : username }}
```
varmar05: I am bit confused - is fullname optinal or always set with username as fallback (from api response)?

server/mergin/sync/models.py
```diff
@@ -364,6 +365,7 @@ class ProjectMember:
     workspace_role: WorkspaceRole
     project_role: Optional[ProjectRole]
     role: ProjectRole
+    fullname: str
```
MarcelGeo: i think use unified value `name` here

MarcelGeo (approved):

PR #558, opened 2026-01-14
V2 list projects endpoint

harminius:
Resolves https://github.com/MerginMaps/server-private/issues/3154

Example:
Screenshot from 2026-01-13 15-47-58

### DIFF
--- a/server/mergin/sync/public_api_v2_controller.py
+++ b/server/mergin/sync/public_api_v2_controller.py
@@ -414,7 +414,7 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
     :param q: Filter by name with ilike pattern
     :type q: str
 
-    :rtype: Dict[str: List[Project], str: Integer, str: String, str: String]
+    :rtype: Dict[str: List[Project], str: Integer, str: Integer, str: Integer]
     """
     ws = current_app.ws_handler.get(workspace_id)
     if not (ws and ws.is_active):


server/mergin/tests/test_public_api_v2.py
```diff
@@ -22,10 +22,10 @@
 from sqlalchemy.exc import IntegrityError
 import pytest
 from datetime import datetime, timedelta, timezone
+import json
```
MarcelGeo, line 25: missing?

server/mergin/sync/workspace.py
```diff
@@ -77,7 +77,9 @@ def user_has_permissions(self, user, permissions):
         if role is WorkspaceRole.OWNER:
             return True
 
-        if permissions == "read":
+        if permissions == "guest":
```
MarcelGeo, line 80: why this? This could not affect rest of endpoints?

server/mergin/sync/public_api_v2.yaml
```diff
@@ -594,6 +668,24 @@ components:
                     format: date-time
                     description: File modification timestamp
                     example: 2024-11-19T13:50:00Z
+    ProjectDetail:
```
MarcelGeo, line 671: Why do we need new ProjectDetail?

server/mergin/tests/test_public_api_v2.py
harminius: json lib is useful for loading response data
`resp_data = json.loads(resp.data)`
you can examine the response in debug mode - without saving the response to the variable I see a loading values for the object in my IDE, which is inconvenient.

server/mergin/sync/workspace.py
harminius: IMO it should be there there in oder to be able to share a specific project with a user when all global roles are disabled
in that case, the user is a guest, and without this update, this method will return false, and the access would be refused

yes, need to be careful about this, I can double-check the implementations of this method

server/mergin/sync/public_api_v2.yaml
harminius: it's Project with files
- in the list project endpoint I want to return Project without files so I separated those two

server/mergin/sync/public_api_v2.yaml
```diff
@@ -367,6 +367,73 @@ paths:
...
+            type: integer
+            maximum: 50
+            example: 50
+        - name: order_params
+          in: query
+          description: Sorting fields e.g. "name ASC,udpdated DESC"
+          required: false
+          schema:
+            type: string
+            example: name_asc,updated_desc
```
varmar05: hm, description and example does not match, I remember there are two ways, just do not know which one is newer (and should be preferred)

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -396,3 +396,49 @@ def upload_chunk(id: str):
...
+    :param page: page number
+    :type page: int
+    :param per_page: Number of results per page
+    :type per_page: int
+    :param order_params: Sorting fields e.g. "name ASC,updated DESC"
+    :type order_params: str
+    :param q: Filter by name with ilike pattern
+    :type q: str
+
+    :rtype: Dict[str: List[Project], str: Integer]
```
varmar05: there are more keys in output

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -396,3 +396,49 @@ def upload_chunk(id: str):
         UploadChunkSchema().dump({"id": chunk_id, "valid_until": valid_until}),
         200,
     )
+
+
+@auth_required
+def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=None):
+    """Paginate over workspace projects with optional filtering.
+
+    :param page: page number
```
varmar05, line 408: workspace id is missing

server/mergin/sync/workspace.py
```diff
@@ -77,7 +77,9 @@ def user_has_permissions(self, user, permissions):
...
+            return role >= WorkspaceRole.GUEST
```
varmar05: just wondering if we should not test equality here as 'guest' is a bit different - I do not think reader inherits guest as such

server/mergin/tests/test_public_api_v2.py
varmar05: hm, sometimes we use response.json and sometimes loads(resp.data) - it would be good to be consistent and I would vote for better performance (if any) due to automated tests in CI

server/mergin/tests/test_public_api_v2.py
```diff
@@ -598,3 +599,81 @@ def test_full_push(client):
...
+    response = client.get(url + f"?page={page}&per_page={per_page}")
+    assert response.json["count"] == 11
+    assert len(response.json["projects"]) == 2
+    # name search - more results
+    page = 1
+    per_page = 3
+    response = client.get(url + f"?page={page}&per_page={per_page}&q=1")
+    assert response.json["count"] == 2
+    assert len(response.json["projects"]) == 2
+    assert response.json["projects"][1]["name"] == "project_10"
```
varmar05, line 636: I would be bit careful here and pass also explicit order parameter (or at least make a comment)

server/mergin/tests/test_public_api_v2.py
```diff
@@ -598,3 +599,81 @@ def test_full_push(client):
...
+    # name search - specific result
+    project_name = "project_4"
+    response = client.get(url + f"?page={page}&per_page={per_page}&q={project_name}")
+    assert response.json["projects"][0]["name"] == project_name
+
+    # no permissions to workspace
+    user2 = add_user("user", "password")
+    login(client, user2.username, "password")
+    Configuration.GLOBAL_READ = 0
```
varmar05: Would not this affect subsequent tests? I'd prefer mock the value.

server/mergin/sync/workspace.py
harminius: :heavy_check_mark:

server/mergin/tests/test_public_api_v2.py
harminius: :heavy_check_mark:

server/mergin/tests/test_public_api_v2.py
harminius: :heavy_check_mark:

server/mergin/tests/test_public_api_v2.py
harminius: the performance should be identical - response lib uses `json.loads` internally, and it caches the result

server/mergin/sync/public_api_v2_controller.py
harminius: :heavy_check_mark:

server/mergin/sync/public_api_v2_controller.py
harminius: :heavy_check_mark:

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -412,7 +414,7 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
     :param q: Filter by name with ilike pattern
     :type q: str
 
-    :rtype: Dict[str: List[Project], str: Integer]
+    :rtype: Dict[str: List[Project], str: Integer, str: String, str: String]
```
varmar05: are not page / per page integers?

harminius: yes, sorry

server/mergin/sync/workspace.py
MarcelGeo: what's result from this research?

harminius: `user_has permission` method with `guest` permission (`ws.user_has_permissions(current_user, "guest")`) is used only at this place in CE - other implementations use different permissions - so nothing can be affected by adding `guest` permission to the method

MarcelGeo (approved):

PR #565, opened 2026-01-22
Workspaces API - order by schema fields instead of database fields

harminius:
Fixes 1. and 3. in https://github.com/MerginMaps/server-private/issues/3154#issuecomment-3778290318

- adds helper to map API field names to DB column names based on Schema
- when invalid order param is passed error is thrown

### DIFF
--- a/server/mergin/sync/public_api_v2_controller.py
+++ b/server/mergin/sync/public_api_v2_controller.py
@@ -443,7 +443,7 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
         )
         projects = projects.order_by(*order_by_params)
 
-    pagination = projects.paginate(page, per_page)
+    pagination = projects.paginate(page=page, per_page=per_page)
     result = pagination.items
     total = pagination.total
 
--- a/server/mergin/utils.py
+++ b/server/mergin/utils.py
@@ -162,7 +162,9 @@ def get_schema_fields_map(schema: Type[Schema]) -> dict:
         if not field:
             continue
         # skip virtual fields as DB cannot sort by them
-        if isinstance(field, (fields.Function, fields.Method)):
+        if isinstance(
+            field, (fields.Function, fields.Method, fields.Nested, fields.List)
+        ):
             continue
         if field.attribute:
             mapping[name] = field.attribute


server/mergin/utils.py
```diff
@@ -43,16 +44,22 @@ def get_order_param(
...
+    if "." in db_column_name:
+        col, attr = db_column_name.split(".")
     else:
-        col = order_param.name
+        col = db_column_name
         attr = None
     order_attr = cls.__table__.c.get(col, None)
     if not isinstance(order_attr, Column):
-        return
+        abort(400, "Invalid order parameter")
```
MarcelGeo: give it to logs just. We suggest with @varmar05 not use this.

server/mergin/utils.py
```diff
@@ -135,3 +149,16 @@ def save_diagnostic_log_file(app: str, username: str, body: bytes) -> str:
...
+def get_schema_fields_map(schema: Schema) -> dict:
+    """
+    Creates a mapping of schema field names to corresponding DB columns.
+    This allows sorting by the API field name (e.g. 'size') while
+    actually sorting by the database column (e.g. 'disk_usage').
+    """
+    mapping = {}
+    for name, field in schema._declared_fields.items():
+        if field and field.attribute:
+            mapping[name] = field.attribute
```
MarcelGeo, line 170: What about fallback here?

server/mergin/sync/schemas_v2.py
```diff
@@ -46,3 +47,6 @@ class Meta:
             "workspace",
             "role",
         )
+
+
+ProjectSchema.field_map = get_schema_fields_map(ProjectSchema)
```
MarcelGeo: this is necessary? Its not better to have function where this schema will go as param?

harminius: Could be (I had it in my first version). But it seemed cleaner to me to have it here. Moreover, the map is created once and then the "cached" version is used for all API calls.

MarcelGeo: this field_map is like standard or just patched value? What about add it to some method or something instead of here? I think we do not need to cache that.

varmar05: I would not complicate it (by extending marshmallow schemas) and use just plain functions where needed as we are slowly moving away from marshmallow anyway.

server/mergin/utils.py
varmar05: let's have a complete map here even if attribute is the same as db column

server/mergin/tests/test_public_api_v2.py
```diff
@@ -651,11 +651,6 @@ def test_list_workspace_projects(client):
     resp_data = json.loads(response.data)
     assert resp_data["projects"][0]["name"] == project_name
 
-    # invalid order param
-    response = client.get(url + f"?page=1&per_page=10&order_params=invalid DESC")
```
varmar05: maybe let's keep the test, just verify it returns 200

server/mergin/utils.py
varmar05: also would be good to write simple dedicated test for this util function

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -437,11 +437,15 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
-        order_by_params = parse_order_params(Project, order_params)
+        schema_map = get_schema_fields_map(ProjectSchemaV2)
+        order_by_params = parse_order_params(
+            Project, order_params, field_map=schema_map
+        )
         projects = projects.order_by(*order_by_params)
 
-    result = projects.paginate(page, per_page).items
-    total = projects.paginate(page, per_page).total
+    pagination = projects.paginate(page, per_page)
```
varmar05: Could you pass these as kwars? This usually breaks sqla 2.x api.

server/mergin/utils.py
```diff
@@ -135,3 +148,25 @@ def save_diagnostic_log_file(app: str, username: str, body: bytes) -> str:
...
+    This allows sorting by the API field name (e.g. 'size') while
+    actually sorting by the database column (e.g. 'disk_usage').
+    """
+    mapping = {}
+    for name, field in schema._declared_fields.items():
+        # some fields could have been overridden with None to be excluded
+        if not field:
+            continue
+        # skip virtual fields as DB cannot sort by them
+        if isinstance(field, (fields.Function, fields.Method)):
```
varmar05: Shall we also ignore fields.Nested?

PR #566, opened 2026-01-26
Update light theme fonts line heights

xkello:
Update line heights according to design system: https://github.com/MerginMaps/server-private/issues/3181


web-app/packages/lib/src/assets/sass/themes/mm-theme-light/_fonts.scss
```diff
@@ -20,89 +20,102 @@
 
 body {
   font-family: Inter, sans-serif;
+  //line-height: 1;
```
MarcelGeo: remove

PR #567, opened 2026-01-26
Update batch API endpoint, openapi documentation and add batch test

xkello:
Work on ticket: https://github.com/MerginMaps/server-private/issues/3155
-from parent: https://github.com/MerginMaps/product/issues/86

### DIFF
--- a/server/mergin/sync/public_api_v2_controller.py
+++ b/server/mergin/sync/public_api_v2_controller.py
@@ -459,16 +459,15 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
     return jsonify(projects=data, count=total, page=page, per_page=per_page), 200
 
 
-def list_batch_projects(ids):
+def list_batch_projects(body):
     """List projects by given list of UUIDs. Limit to 100 projects per request.
 
     :param ids: List of project UUIDs
     :type ids: List[str]
     :rtype: Dict[str: List[Project]]
     """
-
+    ids = list(dict.fromkeys(body.get("ids", [])))
     # remove duplicates while preserving the order
-    ids = list(dict.fromkeys(ids))
     max_batch = current_app.config.get("MAX_BATCH_SIZE", 100)
     if len(ids) > max_batch:
         return BatchLimitError().response(400)


server/mergin/sync/public_api_v2.yaml
```diff
@@ -367,6 +367,77 @@ paths:
...
+              type: object
+              required: [ids]
+              properties:
+                ids:
+                  type: array
+                  minItems: 1
+                  maxItems: 100
+                  description: List of project UUIDs to fetch
+                  items:
+                    type: string
```
harminius: We can use `ProjectId` parameter defined in this file later `v`

server/mergin/sync/public_api_v2.yaml
```diff
@@ -367,6 +367,77 @@ paths:
...
+                    format: uuid
+            examples:
```
harminius: not sure if we need another `examples` if we reference schema, where is an example for each field.. It also apllies for another `examples` beneath.

server/mergin/sync/public_api_v2.yaml
```diff
@@ -547,6 +618,19 @@ components:
       example:
         code: UploadError
         detail: "Project version could not be created (UploadError)"
+    BatchError:
+      type: object
+      properties:
+        id:
```
harminius, line 602: we can again reference the already defined ProjectId

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -437,15 +441,32 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
     return jsonify(projects=data, count=total, page=page, per_page=per_page), 200
+
+@auth_required
+def list_batch_projects():
+    """List projects by given list of UUIDs. Limit to 100 projects per request.
+
+    :rtype: Dict[str: List[Project]]
+    """
+    ids = request.json.get("ids", [])
+    if not ids:
```
harminius: the schema definition already validates presence of `ids`

server/mergin/sync/public_api_v2.yaml
```diff
@@ -367,6 +367,77 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              type: object
+              required: [ids]
+              properties:
+                ids:
+                  type: array
+                  minItems: 1
+                  maxItems: 100
```
harminius: if we want a custom error for this, we should not include the `maxItems` limit here - I believe breaking this validation raises somegeneric error, and the request does not reach the check in the controller

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -437,15 +441,32 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
+        abort(400, "No project UUIDs provided")
+    if len(ids) > 100:
+        abort(400, "BatchLimitExceeded")
```
harminius: not sure if we want a custom error structure

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -445,3 +449,24 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
```
MarcelGeo: Use standard error code response. E.g.

```
code: BatchLimitExceeded
detail: 'blah'
```

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -445,3 +449,24 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
+
+    items = require_project_by_many_uuids(ids, ProjectPermissions.Read, expose=False)
+    project_objects = [i for i in items if not isinstance(i, dict)]
+    error_items  = [i for i in items if isinstance(i, dict)]
+
+    project_data = ProjectSchemaV2(many=True).dump(project_objects)
```
MarcelGeo: Is here provided the same order as in requested `ids` array? I think the order will be lost here.

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,60 @@ def require_project_by_uuid(
 
     return project
 
+def require_project_by_many_uuids(
+    uuids: list[str], permission: ProjectPermissions, scheduled=False, expose=True
```
harminius: IMO: I would follow the task description and not complicate the function with the parameter. There is only one specific use case for this endpoint. We can extend it when another demand comes in.

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -445,3 +449,24 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
+    :rtype: Dict[str: List[Project]]
+    """
+    ids = request.json.get("ids", [])
+    if not ids:
+        abort(400, "No project UUIDs provided")
+    if len(ids) > 100:
+        abort(400, "BatchLimitExceeded")
+
+    items = require_project_by_many_uuids(ids, ProjectPermissions.Read, expose=False)
+    project_objects = [i for i in items if not isinstance(i, dict)]
```
MarcelGeo: We probably do not need this transfomation. We can send it as result of function, or?

server/mergin/tests/test_public_api_v2.py
```diff
@@ -693,3 +693,67 @@ def test_list_workspace_projects(client):
     # logout
     logout(client)
     assert client.get(url + "?page=1&per_page=10").status_code == 401
+
+
+
+
+def test_list_projects_in_batch_access_table(client):
```
MarcelGeo: add also tests for new introduced function in `permissions.py`.

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,60 @@ def require_project_by_uuid(
...
+        permission (ProjectPermissions): The permission level required to access the projects.
+        scheduled (bool, optional): If ``True``, bypasses the check for projects marked for deletion.
+        expose (bool, optional): Controls security disclosure behavior on permission failure.
+            - If `True`: Returns 403 Forbidden (reveals project exists but access is denied).
+            - If `False`: Returns 404 Not Found (hides project existence for security).
+            Standard is that reading results in 404, while writing results in 403
+    """
+    valid_uuids = [uuid for uuid in uuids if is_valid_uuid(uuid)]
+    if not valid_uuids:
+        abort(404)
```
MarcelGeo: I think every item in uuids needs to be valid uuid. If one item is bad here, raise 400 (not 404)

server/mergin/tests/test_public_api_v2.py
```diff
@@ -704,3 +693,67 @@ def test_list_workspace_projects(client):
...
+    p = Project.query.get(public_proj.id)
+    p.public = True
+    db.session.commit()
+
+    url = "/v2/projects/batch"
+    priv_id = str(private_proj.id)
+    pub_id = str(public_proj.id)
+
+    # must be logged in (matches suite behavior)
+    login(client, DEFAULT_USER[0], DEFAULT_USER[1])
```
harminius: admin is logged in by default, I believe

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,60 @@ def require_project_by_uuid(
...
+) -> list[dict]:
+    """
+    Retrieves multiple projects by their UUIDs after validating existence, workspace status, and permissions.
+
+    Args:
+        uuids (list[str]): The unique identifiers of the projects.
+        permission (ProjectPermissions): The permission level required to access the projects.
+        scheduled (bool, optional): If ``True``, bypasses the check for projects marked for deletion.
+        expose (bool, optional): Controls security disclosure behavior on permission failure.
+            - If `True`: Returns 403 Forbidden (reveals project exists but access is denied).
```
MarcelGeo: Do we need this options `expose` and `scheduled`?

server/mergin/sync/public_api_v2.yaml
```diff
@@ -367,6 +367,77 @@ paths:
...
+                          id: 123
+                          name: mergin
+                        role: reader
+                      - id: 9b2b1a2b-3e73-4f1c-9f7a-1a2b3c4d5e6f
+                        error: 404
+        "400":
+          $ref: "#/components/responses/BadRequest"
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "403":
```
harminius: is there really a scenario for `403` response?

server/mergin/sync/public_api_v2_controller.py
harminius: I think this doesn't transform anything. This is just used to decide what to pass to Project schema and what to BatchError schema

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,60 @@ def require_project_by_uuid(
...
+    if not valid_uuids:
+        abort(404)
+
+    projects = Project.query.filter(Project.id.in_(valid_uuids)).filter(
+        Project.storage_params.isnot(None)
+    )
+    if not scheduled:
+        projects = projects.filter(Project.removed_at.is_(None))
+    projects = projects.all()
+    if not projects:
```
MarcelGeo: If project is not found, it needs to be in rresponse with error key.

server/mergin/sync/public_api_v2.yaml
xkello: I went with the table from ticket (https://github.com/MerginMaps/product/issues/86), where in a scenario User is logged in, doesn't have access and project is not public, response is error, 403

server/mergin/sync/public_api_v2.yaml
xkello: Apparently this depends on the way in which the app runs. But yes in our case this might be an issue. Removing it

server/mergin/sync/public_api_v2.yaml
harminius: 1. Two types of responses are mixed here. `responses` defined here are for the whole request, not individual project permission. I mean, if you don't have access to a project, you get `200` response code with the `403` error defined in the response body.
2. I didn't find the place in the code with `abort(403)` - it is what is meant by response definition in this file

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,60 @@ def require_project_by_uuid(
...
+        abort(404)
+
+    filtered_projects = []
+    for project in projects:
+        
+        workspace = project.workspace
+        if not workspace:
+            continue
+        if not is_active_workspace(workspace):
```
harminius: I'd prefer to add a `404` error for this use case

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,60 @@ def require_project_by_uuid(
...
+                # we don't want to tell anonymous user if a private project exists
+                filtered_projects.append({"id": project.id, "error": 404})
+                continue
+            # logged in - YES, have access - NO, public project - NO
+            elif not current_user.is_anonymous and not project.public:
+                filtered_projects.append({"id": project.id, "error": 403})
+                continue
+        filtered_projects.append(project)
+
+    if not filtered_projects:
```
harminius: maybe it's fine to return the empty list here..

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,60 @@ def require_project_by_uuid(
...
+    """
+    valid_uuids = [uuid for uuid in uuids if is_valid_uuid(uuid)]
+    if not valid_uuids:
+        abort(404)
+
+    projects = Project.query.filter(Project.id.in_(valid_uuids)).filter(
+        Project.storage_params.isnot(None)
+    )
+    if not scheduled:
+        projects = projects.filter(Project.removed_at.is_(None))
```
harminius: I think clients want to see these remove files with `404` error in the response

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,60 @@ def require_project_by_uuid(
...
+    projects = projects.all()
+    if not projects:
+        abort(404)
```
harminius: I think there should be no abort if there are valid uuids.
My preference would be even if there is a single valid uuid, but that's a matter of opinion.

server/mergin/sync/permissions.py
```diff
@@ -248,57 +248,53 @@ def require_project_by_uuid(
     return project
 
 def require_project_by_many_uuids(
-    uuids: list[str], permission: ProjectPermissions, scheduled=False, expose=True
+    uuids: list[str], permission: ProjectPermissions
 ) -> list[dict]:
```
harminius: would be nice if it returned `list[Project]`

server/mergin/sync/permissions.py
```diff
@@ -248,57 +248,53 @@ def require_project_by_uuid(
...
+                    else abort(400, {"code" : "InvalidUUID", "detail" : "Invalid UUID format"}) 
+                    for uuid in uuids
+                ]
 
-    projects = Project.query.filter(Project.id.in_(valid_uuids)).filter(
-        Project.storage_params.isnot(None)
+    projects = (
+        Project.query
+        .filter(Project.id.in_(valid_uuids))
+        .filter(Project.storage_params.isnot(None))
```
harminius: i think also these should be included in the response with `404` error

server/mergin/tests/test_permissions.py
```diff
@@ -116,3 +129,67 @@ def test_project_permissions(client):
...
+    p.public = True
+    db.session.commit()
+
+    priv_id = str(private_proj.id)
+    pub_id = str(public_proj.id)
+
+    # First user with access to both projects
+    login(client, DEFAULT_USER[0], DEFAULT_USER[1])
+
+    with client:
```
harminius, line 153: Hmm.. this is to pass current user?

server/mergin/tests/test_public_api_v2.py
```diff
@@ -714,46 +710,19 @@ def test_list_projects_in_batch_access_table(client):
     priv_id = str(private_proj.id)
     pub_id = str(public_proj.id)
 
-    # must be logged in (matches suite behavior)
-    login(client, DEFAULT_USER[0], DEFAULT_USER[1])
+    #login(client, DEFAULT_USER[0], DEFAULT_USER[1]) TODO: why needed?
```
harminius: rm

server/mergin/sync/public_api_v2.yaml
```diff
@@ -367,6 +367,53 @@ paths:
...
+          application/json:
+            schema:
+              type: object
+              required: [ids]
+              properties:
+                ids:
+                  type: array
+                  description: List of project UUIDs to fetch
+                  items:
+                    $ref: "#/components/parameters/ProjectId"
```
harminius: It's probably due to my previous comment - sorry for that, but here is the **path** parameter referenced
A clean solution would be to decouple the definition from the location - to move the definition under the Schema component and put the reference in the Parameters components. 
Then you can safely reference the schema here in the request body definition

```
components:
  parameters:
    ProjectIdParam:
      name: id
      in: path
      required: true
      schema:
        $ref: '#/components/schemas/ProjectIdSchema'
```

server/mergin/sync/public_api_v2.yaml
```diff
@@ -547,6 +594,17 @@ components:
       example:
         code: UploadError
         detail: "Project version could not be created (UploadError)"
+    BatchItemError:
+      type: object
+      properties:
+        id:
+          $ref: "#/components/parameters/ProjectId"
```
harminius: the same previous comment - reference a ProjectId schema here, not parameter

harminius (changes requested):
Great job, we're almost there! 👍 

I have only one major comment. 
I have doubts about the usefulness of `require_project_by_many_uuids()`. At least it should return a list of projects, and the rest (`by_id` dict creation) should be done in the controller. However, since this function needs to list even removed projects, it doesn't seem reusable.
The idea of the help function is that it can be used in a different use case, while this function is designed for this specific usecase.

server/mergin/tests/test_permissions.py
xkello: Yes and no, the line under is also needed. without these 2 test failure was returning error where user was none. I believe it has something to do with active request context.

server/mergin/sync/permissions.py
xkello: Cant be done I believe since we are returning Project objects as well as per UUID errors. The Logic would have to be split somehow

server/mergin/sync/permissions.py
xkello: This is already handled at line 277-284. When uuid is not present in project lists it returns 404 automatically regardless of the reason. So in a way the information why was 404 returned is lost but I dont see a reason to store that info since we dont use it - just a side note though

harminius: I see. I'd also filter out `Project.removed_at.is_(None)` then.

server/mergin/sync/permissions.py
harminius: Dict juggling on lines 272 to 296 can nicely be done in the controller.
Now I see it can also filter out those removed projects which is desirable.
Just a bit of code moving, no logic change to make the output of this function generic and reusable.

server/mergin/sync/permissions.py
```diff
@@ -247,6 +247,56 @@ def require_project_by_uuid(
...
+            continue
+
+        workspace = project.workspace
+        if not workspace or not is_active_workspace(workspace):
+            filtered_projects.append({"id": uuid, "error": 404})
+            continue
+
+        if not permission.check(project, current_user):
+            # logged in - NO, have acccess - NONE, public project - NO
+            if current_user.is_anonymous and not project.public:
```
harminius: does `not project.public` condition makes sense?
if the project is public, the `if` permission check above will be `True`..

server/mergin/tests/test_public_api_v2.py
```diff
@@ -704,3 +693,61 @@ def test_list_workspace_projects(client):
...
+    assert "projects" in resp.json
+    assert isinstance(resp.json["projects"], list)
+    assert len(resp.json["projects"]) == 2
+    # Both projects returned as full objects for admin
+    for proj in resp.json["projects"]:
+        assert "id" in proj
+        assert "name" in proj  # full project object
+
+    # Reset global permissions
+    from ..config import Configuration
```
harminius: I suppose import is not needed here

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -437,15 +447,59 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
-    pagination = projects.paginate(page=page, per_page=per_page)
-    result = pagination.items
-    total = pagination.total
+    result = projects.paginate(page, per_page).items
+    total = projects.paginate(page, per_page).total
 
     data = ProjectSchemaV2(many=True).dump(result)
     return jsonify(projects=data, count=total, page=page, per_page=per_page), 200
+
+@auth_required
```
harminius: i think we don't want this, best to check with @varmar05

server/mergin/tests/test_public_api_v2.py
```diff
@@ -704,3 +693,61 @@ def test_list_workspace_projects(client):
     # logout
     logout(client)
     assert client.get(url + "?page=1&per_page=10").status_code == 401
+
+def test_list_projects_in_batch(client):
```
harminius, line 699: can we test ivalid uuid?

xkello: Well we were able to but it got lost somewhere in the changes... Adding it

server/mergin/sync/public_api_v2_controller.py
xkello: Yes I too believe so. Since anonymous users can also view public projects. But lets wait for reply from Martin

varmar05: yes, let's remove it - those getters for projects do not have auth decorator due to public projects

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -437,15 +447,58 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
     return jsonify(projects=data, count=total, page=page, per_page=per_page), 200
+
+def list_batch_projects():
+    """List projects by given list of UUIDs. Limit to 100 projects per request.
+
+    :rtype: Dict[str: List[Project]]
+    """
+    ids = request.json.get("ids", [])
+    if len(ids) > 100:
+        abort(400,
```
varmar05: Shall we create a custom error code class for this?

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -437,15 +447,58 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
+    ids = request.json.get("ids", [])
+    if len(ids) > 100:
+        abort(400, 
+            {
+                "code" : "BatchLimitExceeded",
+                "detail" : "A maximum of 100 project UUIDS can be requested at once"
+            }
+        )
+
+    valid_uuids = [uuid
```
varmar05: I would rather do some assert and only abort if failed - it would take bit longer but a code clearer without hidden abort.

Again, I am not sure if we need / want to have custom error - but we should stick to general custom error framework.

server/mergin/tests/test_public_api_v2.py
```diff
@@ -704,3 +693,75 @@ def test_list_workspace_projects(client):
...
+    p = Project.query.get(public_proj.id)
+    p.public = True
+    db.session.commit()
+
+    url = "/v2/projects/batch"
+    priv_id = str(private_proj.id)
+    pub_id = str(public_proj.id)
+
+    # missing ids -> 400 (connexion validation)
+    resp = client.post(url, json={})
```
varmar05, line 716: what about testing length check?

server/mergin/tests/test_permissions.py
```diff
@@ -116,3 +129,48 @@ def test_project_permissions(client):
...
+        client.get("/")
+        assert check_project_permissions(priv_proj, ProjectPermissions.Read) is None
+        assert check_project_permissions(pub_proj, ProjectPermissions.Read) is None
+
+    # Reset global permissions for subsequent tests
+    Configuration.GLOBAL_READ = False
```
varmar05: I think we should avoid changing configuration and use instead mocking for particular use case - to avoid effect on other tests.

server/mergin/sync/permissions.py
```diff
@@ -248,6 +248,32 @@ def require_project_by_uuid(
...
+    :param project: project
+    :type project: Project
+    :param permission: permission to check
+    :type permission: ProjectPermissions
+    :return: HTTP error code if permission check fails, None otherwise
+    :rtype: int | None
+    """
+    
+    workspace = project.workspace
+    if not workspace or not is_active_workspace(workspace):
```
varmar05: I am bit concerned this is called in a loop and will do many selects. It would be better to join workspace and its active status in a project query. But we would need to be more cautious here as it works differently in CE and EE.

What do you think @MarcelGeo?

server/mergin/sync/public_api_v2_controller.py
MarcelGeo: no , just use abort(400)

server/mergin/tests/test_public_api_v2.py
```diff
@@ -768,3 +769,10 @@ def test_list_projects_in_batch(client):
     # private -> error 404 (anonymous cannot access private)
     priv_result = next(p for p in projects if p.get("id") == priv_id)
     assert priv_result["error"] == 404
+
+    # batch size limit: generate more than allowed uuids and expect error
+    max_batch = client.application.config.get("MAX_BATCH_SIZE", 100)
```
harminius, line 774: if it is config, we could also mock config value to `1` but it's not a big deal

xkello: Thought about this case, but I couldn't think of any cases where length 1 could be a problem

harminius (approved):

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -437,15 +449,47 @@ def list_workspace_projects(workspace_id, page, per_page, order_params=None, q=N
...
+
+
+def list_batch_projects():
+    """List projects by given list of UUIDs. Limit to 100 projects per request.
+
+    :rtype: Dict[str: List[Project]]
+    """
+
+    # remove duplicates while preserving the order
+    ids = list(dict.fromkeys(request.json.get("ids", [])))
```
varmar05: ids should be passed by connexion lib as function argument

PR #593, opened 2026-03-13
Handle nullable latest_file_history_ids

varmar05:
This prevent missing item in project version files when list of history ids has not been cached yet.

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -135,7 +135,7 @@ class Project(db.Model):
         project_workspace = current_app.ws_handler.get(self.workspace_id)
         return project_workspace
 
-    def get_latest_file_history_ids(self) -> List[int]:
+    def get_latest_files_cache(self) -> List[int]:
         """Get latest file history ids either from cached table or calculate them on the fly"""
         if self.latest_project_files.file_history_ids is not None:
             return self.latest_project_files.file_history_ids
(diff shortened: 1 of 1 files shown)


server/mergin/sync/models.py
```diff
@@ -135,6 +135,36 @@ def workspace(self):
...
+            WITH latest_changes AS (
+                SELECT
+                    fp.id,
+                    pv.project_id,
+                    max(pv.name) AS version
+                FROM
+                    project_version pv
+                    LEFT OUTER JOIN file_history fh ON fh.version_id = pv.id
+                    LEFT OUTER JOIN project_file_path fp ON fp.id = fh.file_path_id
+                WHERE
```
MarcelGeo, line 153: No order by here?

server/mergin/sync/models.py
```diff
@@ -1434,7 +1468,7 @@ def __init__(
         latest_files_map = {
             fh.path: fh.id
             for fh in FileHistory.query.filter(
-                FileHistory.id.in_(self.project.latest_project_files.file_history_ids)
+                FileHistory.id.in_(self.project.get_latest_file_history_ids())
```
MarcelGeo: Something like a get_latest_files_cache() ?

MarcelGeo (changes requested):

server/mergin/sync/models.py
varmar05: we use max here for later use in join - the whole query could be written using distinct and order by but I wanted it to stay similar we already have  in cache_latest_files()

MarcelGeo (approved):

PR #594, opened 2026-03-13
For gpkg file restore always use diff checkpoints

varmar05:
Drop legacy support to use individual 0 rank diffs. Instead always use diff checkpoints and create them recursively if needed. This change uses the same logic as we do in v2 endpoints.

Please note that reconstruction can still throw 502/504 if backend does not make it within allowed time. As each try can now take less time to finish (intermediate checkpoints created), clients can implement retry logic.

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -743,22 +743,21 @@ class FileHistory(db.Model):
             return None, []
 
         diffs = []
-        cached_items = Checkpoint.get_checkpoints(
-            basefile.project_version_name, version
-        )
+        checkpoints = Checkpoint.get_checkpoints(basefile.project_version_name, version)
         expected_diffs = (
             FileDiff.query.filter_by(
                 basefile_id=basefile.id,
             )
             .filter(
                 tuple_(FileDiff.rank, FileDiff.version).in_(
-                    [(item.rank, item.end) for item in cached_items]
+                    [(item.rank, item.end) for item in checkpoints]
                 )
             )
             .all()
         )
 
-        for item in cached_items:
+        for item in checkpoints:
+            diff_needs_to_be_created = False
             diff = next(
                 (
                     d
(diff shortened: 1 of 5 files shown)


server/mergin/sync/public_api_controller.py
```diff
@@ -337,16 +337,23 @@ def download_project_file(
...
-            file, ProjectVersion.from_v_name(version)
-        )
-
     abs_path = os.path.join(project.storage.project_dir, file_path)
-    # check file exists (e.g. there might have been issue with restore)
+
     if not os.path.exists(abs_path):
-        logging.error(f"Missing file {namespace}/{project_name}/{file_path}")
-        abort(404)
+        if version and not diff:
```
MarcelGeo: Probably add some reolving map on the start of controller, where we can make actions based on this version and diff attributes. 

If not version and diff: abort
If version and diff: download_diff
If not diff and version: dowload_geopackage

Where is handled a non gpkg file?

server/mergin/sync/models.py
```diff
@@ -767,25 +766,38 @@ def diffs_chain(
...
+                        basefile=basefile,
+                        version=item.end,
+                        rank=item.rank,
+                        path=basefile.file.generate_diff_name(),
+                        size=None,
+                        checksum=None,
                     )
-                    .filter(
-                        FileDiff.version >= item.start, FileDiff.version <= item.end
+                    db.session.add(diff)
```
MarcelGeo, line 785: Do we want to follow commit in the end?

server/mergin/tests/test_public_api_v2.py
```diff
@@ -304,45 +304,31 @@ def test_create_diff_checkpoint(diff_project):
...
-    diff = FileDiff(
-        basefile=basefile, path=f"test.gpkg-diff-{uuid.uuid4()}", version=20, rank=1
-    )
-    db.session.add(diff)
-    db.session.commit()
-    assert not os.path.exists(diff.abs_path)
+    diff = FileDiff.query.filter_by(
+        file_path_id=file_path_id, version=20, rank=1
+    ).first()
+    os.remove(diff.abs_path)
```
MarcelGeo, line 317: Why removed assert which was before here on this line?

MarcelGeo (changes requested):
Nice

server/mergin/sync/public_api_controller.py
varmar05: non gpkg only check file.location exist, else abort

server/mergin/sync/models.py
varmar05: yes, I would commit sooner then layer, so we don't loose it in case of timeout

server/mergin/tests/test_public_api_v2.py
varmar05: the logic of tests had to be rewritten here as a lot of checkpoints were created in very first call, unlike in previous case where we used only rank 0 to calculate diffs history

this test now tells following: merged diff already exist, but even if we drop the file, it will be successfully recreated

server/mergin/tests/test_public_api_v2.py
```diff
@@ -368,11 +354,12 @@ def test_create_diff_checkpoint(diff_project):
...
         mock.side_effect = GeoDiffLibError
-        diff = FileDiff(
-            basefile=basefile, path=f"test.gpkg-diff-{uuid.uuid4()}", version=16, rank=1
-        )
-        db.session.add(diff)
-        db.session.commit()
+        diff = FileDiff.query.filter_by(
+            file_path_id=file_path_id, version=16, rank=1
+        ).first()
+        if os.path.exists(diff.abs_path):
```
MarcelGeo: add here assert

MarcelGeo (approved):

PR #596, opened 2026-03-17
Upgrade app banner to have rounded boundaries in footer

MarcelGeo:
:rose: FE:

- added rounded boundaries for app-section-banner when footer is used.


server/mergin/sync/public_api_v2_controller.py
```diff
@@ -204,6 +204,10 @@ def get_project(id, files_at_version=None):
     """Get project info. Include list of files at specific version if requested."""
     project = require_project_by_uuid(id, ProjectPermissions.Read, expose=False)
     data = ProjectSchemaV2().dump(project)
+    if request.method == "HEAD":
+        headers = {"X-Mm-Project-Version": project.latest_version}
+        return data, 200, headers
```
varmar05: do we need data here at all?

MarcelGeo: Yes, beacuse of connexion.

varmar05 (approved):

PR #601, opened 2026-03-25
Remove outdated alembic branchpoint

varmar05:
Branching point in community line from few years ago only creates confusion and make upgrades more difficult. Let's remove the old migration, which cleans `heads` and `history` alembic commands.

**Deployment note** 

Because existing deployments might have stamped version where file physically do not exist it would break alembic commands, e.g. `flask db current` will not work. We can't therefore use `stamp --purge` and we need to remove the row using sql
```
docker exec merginmaps-db psql -U postgres -c "DELETE FROM alembic_version WHERE version_num='35af0c8be41e';"
```
Once it is gone, `flask db` commands work as expected.


MarcelGeo (changes requested):
Should we make this as standard migration step? Like remove branch in alembic?

varmar05:
> Should we make this as standard migration step? Like remove branch in alembic?

Unfortunately it cannot be done via alembic if migration file is already gone. And unstamp if file is still present leads to missing upgrade to that head we want to get rid of (it is like catch 22). So best to patch DB manually and have it in migration guide.

MarcelGeo (approved):

PR #605, opened 2026-03-31
Fix issue with gpkg reconstruction from merged diffs

varmar05:
The issue:

In certain cases full gpkg file could not be reconstructed from diff checkpoints as those were incorrect.
Typically it happened when diff checkpoint started at version of basefile (e.g. v1) and during its construction from lower ranks starting diff was ignored leading to invalid checkpoint file.

Imagine situation like this when we wanted to create some higher rank checkpoint:
```
Checkpoint(rank=2, index=1, versions=v1-v16), 
[
Checkpoint(rank=1, index=1, versions=v1-v4), <-- was not created in recursion!
Checkpoint(rank=1, index=2, versions=v5-v8), 
Checkpoint(rank=1, index=3, versions=v9-v12), 	
Checkpoint(rank=0, index=13, versions=v13-v13), 
Checkpoint(rank=0, index=14, versions=v14-v14), 
Checkpoint(rank=0, index=15, versions=v15-v15) 
Checkpoint(rank=0, index=16, versions=v16-v16)
]
```
where first checkpoint in the list was originally ignored.

Fix:

This PR lessens the condition when checkpoint should be ignored only to situation if it spanned over basefile version.

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -1017,8 +1017,9 @@ class FileDiff(db.Model):
         )
 
         for item in cached_items:
-            # basefile is a start of the diff chain
-            if item.start <= basefile.project_version_name:
+            # basefile is a start of the diff chain, item cannot cross over it,
+            # but merged diffs can start with basefile version containing changes since then
+            if item.start < basefile.project_version_name:
                 continue
 
             # find diff in table and on disk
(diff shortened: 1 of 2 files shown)


MarcelGeo (approved):
Wondering If this could not go to master.

varmar05:
@MarcelGeo I think we still have a mishandled use case:

`can_create_checkpoint` can return true but `construct_checkpoint` will fail to create file on disk - and this might be triggered from delta

example:
v10 basefile
v11, v12 diff
checkpoint (v9-v12, rank 1)

it might not be triggered by clients but if it happens, it might fail

PR #607, opened 2026-04-08
Fix condition for creating diff checkpoints

varmar05:
Enforce rule that checkpoint can not contain basefile in its range. 

When creating diff chain / history, always lookup with version higher than basefile. This prevents confusion when it is worth creating a diff checkpoint and ensures we only create merged diffs which can actually be applied.

As a consequence, gpkg will never have checkpoints starting with version 1.

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -777,7 +777,10 @@ class FileHistory(db.Model):
             return None, []
 
         diffs = []
-        checkpoints = Checkpoint.get_checkpoints(basefile.project_version_name, version)
+        # get all checkpoints with diffs after basefile which should be applied
+        checkpoints = Checkpoint.get_checkpoints(
+            basefile.project_version_name + 1, version
+        )
         expected_diffs = (
             FileDiff.query.filter_by(
                 basefile_id=basefile.id,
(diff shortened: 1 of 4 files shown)


server/mergin/sync/models.py
```diff
@@ -922,6 +925,10 @@ def can_create_checkpoint(file_path_id: int, checkpoint: Checkpoint) -> bool:
         if not basefile:
             return False
 
+        # do not create checkpoint if basefile is present in the range as it does not have valid use case
+        if basefile.project_version_name >= checkpoint.start:
```
MarcelGeo, line 929: do we want to use here also checkpoint.end?

MarcelGeo: also in ⬇️ condition.

server/mergin/tests/test_public_api_v2.py
```diff
@@ -320,8 +322,8 @@ def test_create_diff_checkpoint(diff_project):
 
     basefile, diffs = FileHistory.diffs_chain(file_path_id, 20)
     assert basefile.project_version_name == 9
-    # individual diff v12 + (v13-v16) + (v17-v20) as the last one
-    assert len(diffs) == 3
+    # individual diffs v11, v12 + (v13-v16) + (v17-v20) as the last one
```
MarcelGeo, line 325: interesting, that v11 is here. it was not here before?

MarcelGeo (changes requested):

server/mergin/tests/test_public_api_v2.py
varmar05: it was not, and that is also bit scary! but it should be there

server/mergin/sync/models.py
varmar05: what to check with .end?

MarcelGeo (approved):

PR #608, opened 2026-04-09
Implement upload directory creation and lockfile handling

MarcelGeo:
Move lockfile creation to Upload instance creation to prevent `is_active` collisions within multiple uploads.

This could be replaced by upcoming race conditions and lockfile removing.


varmar05:
closing in favor of #610

PR #609, opened 2026-04-11
Change fetch users logic

xkello:
Regarding issue: https://github.com/MerginMaps/server-private/issues/3263.

Users were not displayed in search drop-down on removal without refreshing page due to fetching them from a static array. They were fetched at the beginning and never again. 

After fix they are fetched from an array which gets refreshed on each addition/removal of an user.


web-app/packages/lib/src/modules/user/store.ts
MarcelGeo, line 302: fix lint

MarcelGeo (approved):

PR #610, opened 2026-04-14
rework concurrent upload

varmar05:
This PR contains two fixes / improvements

- Replace lockfile with upload last_ping db column
- Rework concurrent upload using upsert strategy

### DIFF
--- a/server/mergin/sync/models.py
+++ b/server/mergin/sync/models.py
@@ -1808,7 +1808,9 @@ class Upload(db.Model):
     created = db.Column(db.DateTime, default=datetime.utcnow)
     # last ping time to determine if upload is still active
     last_ping = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
-    transaction_id = db.Column(db.String, unique=True, nullable=False, index=True)
+    transaction_id = db.Column(
+        UUID(as_uuid=True), unique=True, nullable=False, index=True
+    )
 
     user = db.relationship("User")
     project = db.relationship(
@@ -1835,7 +1837,7 @@ class Upload(db.Model):
         """Create upload session, it can either create a new record or handover an existing one but with new transaction id
         Old transaction folder is removed and new one is created.
         """
-        now = datetime.now(timezone.utc)
+        now = datetime.now(timezone.utc).replace(tzinfo=None)
         expiration = current_app.config["LOCKFILE_EXPIRATION"]
         new_tx_id = str(uuid.uuid4())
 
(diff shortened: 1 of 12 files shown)


server/mergin/sync/models.py
```diff
@@ -1805,6 +1806,9 @@ class Upload(db.Model):
         db.Integer, db.ForeignKey("user.id", ondelete="CASCADE"), nullable=True
     )
     created = db.Column(db.DateTime, default=datetime.utcnow)
+    # last ping time to determine if upload is still active
+    last_ping = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
+    transaction_id = db.Column(db.String, unique=True, nullable=False, index=True)
```
MarcelGeo: Add UUID type

server/mergin/sync/models.py
```diff
@@ -1822,28 +1826,166 @@ def __init__(self, project: Project, version: int, changes: dict, user_id: int):
...
 
     @property
-    def lockfile(self):
-        return os.path.join(self.upload_dir, "lockfile")
+    def upload_dir(self):
+        return os.path.join(
+            self.project.storage.project_dir, "tmp", self.transaction_id
+        )
 
     def is_active(self):
```
MarcelGeo: Check If we need this

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -296,88 +295,65 @@ def create_project_version(id):
...
+                get_device_id(request),
+            )
+            db.session.add(pv)
+            db.session.add(project)
+            db.session.commit()
+
+            # let's move uploaded files where they are expected to be
+            if to_be_added_files or to_be_updated_files:
+                temp_files_dir = os.path.join(
+                    upload.upload_dir, "files", v_next_version
```
MarcelGeo, line 340: This could not be the same situation as with the previous solution, but in different way:

- somebody will create version with different data in `v{n}` folder?

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -296,88 +295,65 @@ def create_project_version(id):
...
+            logging.info(
+                f"Push finished for project: {project.id}, project version: {v_next_version}."
+            )
+            project_version_created.send(pv)
+            push_finished.send(pv)
     except (
         psycopg2.Error,
         FileNotFoundError,
         IntegrityError,
-        ObjectDeletedError,
```
MarcelGeo, line 377: This could be still the case, no? When somebody will remove Upload , another process is trying to use it (insert, upsert) . Or?

MarcelGeo (changes requested):
We could try to run some concurrent push test If this will really help and do not move race conditions to different segments.

server/mergin/sync/models.py
varmar05: removed as it was racy and redundant check, no need with upsert strategy anymore

server/mergin/sync/public_api_v2_controller.py
varmar05: no, it can't - there is no way for 2 concurrent uploads to be running simultaneously from the same base version with the new upsert strategy

server/mergin/sync/public_api_v2_controller.py
varmar05: it would only happen if row is removed outside of push endpoints (upsert prohibits this), e.g. on delete cascade, but that never happens as we do not remove project or user rows anyway, only anonymize

MarcelGeo (approved):

PR #611, opened 2026-04-14
Merge user_profile into user table

varmar05:
We keep responses untouched, the only change happens on DB / model side to avoid excessive joins.


varmar05:
Note: based on order of merges #610 we need to update alembic migration hash to keep history linear

MarcelGeo (approved):

PR #613, opened 2026-04-28
Fix create diff error

varmar05:
Fix the error for creating diff files on server side which would happen if temp copy and target diff file had the same name and was removed from disk by mistake (e.g. in case of geodiff copy error).

### DIFF
--- a/server/mergin/sync/storages/disk.py
+++ b/server/mergin/sync/storages/disk.py
@@ -333,7 +333,7 @@ class DiskStorage(ProjectStorage):
                 # create changeset next to uploaded file copy
                 changeset_tmp = os.path.join(
                     os.path.dirname(uploaded_file_tmp),
-                    diff_name,
+                    diff_name + "_tmp",
                 )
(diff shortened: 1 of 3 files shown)


MarcelGeo (changes requested):
Do we want to add some guards when project changeset fails?

varmar05:
> Do we want to add some guards when project changeset fails?

What guards do you mean? If changeset fails it will do force update instead.

MarcelGeo:
> > Do we want to add some guards when project changeset fails?
> 
> What guards do you mean? If changeset fails it will do force update instead.

I meant this, If we do not want to add some Os errors handling or it's fine.
We can make it another issue.

```
except (GeoDiffLibError, GeoDiffLibConflictError) as e:
                # diff is not possible to create - file will be overwritten
                move_to_tmp(diff_file)
                return Err(self.gediff_log.getvalue())
```

MarcelGeo (approved):

PR #618, opened 2026-05-04
Add admin panel enhancements according to #3283

xkello:
Add functionality according to https://github.com/MerginMaps/server-private/issues/3283.

1. Users query search now persists when going to profile and back to search.
2. Table is not updated every input now.
3. Users are now able to open new tab/window from the result table

### DIFF
--- a/web-app/packages/admin-lib/src/modules/admin/components/AccountsTable.vue
+++ b/web-app/packages/admin-lib/src/modules/admin/components/AccountsTable.vue
@@ -24,7 +24,7 @@
           <PInputText
             placeholder="Search accounts"
             data-cy="search-members-field"
-            v-model="searchByName"
+            v-model="search"
             class="w-full"
             @input="onSearch"
           />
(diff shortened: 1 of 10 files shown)


web-app/packages/lib/src/common/composables/useDataTableSearch.ts
```diff
@@ -0,0 +1,104 @@
...
+  defaultSortDesc?: boolean
+}
+
+/**
+ * Shared search/pagination/URL-sync logic for lazy-loaded admin data tables.
+ *
+ * Call setFetchFn() immediately after setup to register the table-specific
+ * fetch action; all event handlers will invoke it automatically.
+ */
+export function useDataTableSearch(opts: DataTableSearchOptions = {}) {
```
MarcelGeo, line 22: It's not better to add here also watcher for route query params?

web-app/packages/lib/src/common/composables/useDataTableSearch.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (q.per_page) options.itemsPerPage = Number(q.per_page)
+    if (q.order_by) options.sortBy[0] = String(q.order_by)
+    if (q.desc) options.sortDesc[0] = q.desc === 'true'
+  }
+
+  function updateQuery() {
+    const query: Record<string, string> = {}
+    if (search.value) query.q = search.value
+    if (options.page > 1) query.page = String(options.page)
+    if (options.itemsPerPage !== 20)
```
MarcelGeo: some hardcoded constant? I think we can add it to options too with default 20

web-app/packages/admin-lib/src/modules/admin/components/AccountsTable.vue
```diff
@@ -44,49 +44,45 @@
...
+            <template #body="{ data }">
+              <router-link :to="accountRoute(data)" class="dt-row-link">
+                {{ data.profile?.name }}
+              </router-link>
+            </template>
+          </PColumn>
+          <PColumn field="active" header="Active">
+            <template #body="{ data }">
+              <router-link :to="accountRoute(data)" class="dt-row-link">
+                <i v-if="data.active" class="ti ti-check" />
```
MarcelGeo: Just note from some users. We could add here just text Active / Inactive as icon is not copy able.

web-app/packages/admin-lib/src/modules/admin/components/AccountsTable.vue
```diff
@@ -44,49 +44,45 @@
...
           :sort-order="options.sortDesc[0] ? -1 : 1"
+          :rowHover="true"
           removableSort
           reorderable-columns
           @page="onPage"
-          @row-click="rowClick"
           @sort="onSort"
           data-cy="accounts-table"
         >
-          <template v-for="header in headers" :key="header.field">
```
MarcelGeo, line 54: This means also for other tables. We could probably live with this config json as we can then extend it and have one default column when slot is not used.

MarcelGeo (changes requested):

web-app/packages/lib/src/common/composables/useDataTableSearch.ts
xkello: I think it would cause an infinite loop: doFetch -> updateQuery -> router.replace -> route.query changes -> watcher -> doFetch... A guard flag could break it but it's unnecessary - the component remounts on back navigation so created() already handles state restore. A watcher would only add value in my opinion if the URL could change externally while the component is mounted which isn't the case here.

web-app/packages/admin-lib/src/modules/admin/components/AccountsTable.vue
```diff
@@ -44,48 +44,34 @@
...
-                  :to="{
-                    name: 'account',
-                    params: { username: slotProps.data.username }
-                  }"
+                  :to="accountRoute(data)"
+                  class="dt-row-link"
+                  :class="header.class"
                 >
-                  {{ slotProps.data.username }}
+                  <template v-if="header.type === 'boolean'">{{
```
MarcelGeo: I preffer to look based on field attribute. Probably we can get rid of type boolean column.

MarcelGeo (changes requested):

MarcelGeo (approved):

PR #619, opened 2026-05-05
Handle rename error in case of full disk

varmar05:


server/mergin/sync/storages/disk.py
```diff
@@ -114,7 +114,11 @@ def move_to_tmp(src, dest=None):
             else:
                 root = tempfile.gettempdir()
             temp_path = os.path.join(root, "delete-me-" + dest, os.path.basename(src))
-            os.renames(src, temp_path)
+            try:
+                os.renames(src, temp_path)
+            except OSError as rename_err:
+                logging.error(f"Failed to move {src} to tmp: {rename_err}")
+                return None
```
MarcelGeo, line 121: What happens for sync mechanism, when return is here? Some error? Do we want to propagate this and block somebody from syncing in such a case?

MarcelGeo (approved):

server/mergin/sync/storages/disk.py
varmar05: I would not block sync - this is just a cleanup which should not affect user - in production code return value is ignored anyway (used just in tests).

MarcelGeo: Nice... that's ok when not blocking sync

PR #635, opened 2026-06-01
Add configurable confirm dialog logo and update Project Settings

xkello:
Add configuration to change confirm dialog logo bg to fill the figma (https://www.figma.com/design/Zl8irhQtrHrxBVcJyLPedL/Dashboard-2.0?node-id=3355-10739&t=mtqPIiHeupfZI8d1-0) requirements. Behavior only applies when it is defined for the logo. When its left empty (default setting) the logo follows old logic - bg changes according to button variant. 

Update some names according to figma specs above

### DIFF
--- a/web-app/packages/lib/src/modules/dialog/components/ConfirmDialog.vue
+++ b/web-app/packages/lib/src/modules/dialog/components/ConfirmDialog.vue
@@ -6,17 +6,7 @@ SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
 
 <template>
   <div class="flex flex-column align-items-center pb-4 text-center gap-2">
-    <img
-      v-if="severity === 'danger'"
-      src="@/assets/trash.svg"
-      alt="Cover for confirm dialog"
-    />
-    <img
-      v-else-if="severity === 'warning'"
-      src="@/assets/warning-dialog.svg"
-      alt="Cover for confirm dialog"
-    />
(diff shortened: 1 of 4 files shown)


web-app/packages/lib/src/modules/dialog/components/ConfirmDialog.vue
```diff
@@ -91,6 +86,31 @@ const isConfirmed = computed(() => {
     : true
 })
 
+const logoSrc = computed(() => {
+  const variant =
+    props.logoVariant === 'auto'
```
MarcelGeo, line 91: what means auto here?

web-app/packages/lib/src/modules/project/views/ProjectViewTemplate.vue
```diff
@@ -208,7 +208,7 @@ export default defineComponent({
           })
           tabs.push({
             route: ProjectRouteName.ProjectSettings,
-            header: 'Settings'
+            header: 'Settings & API'
```
MarcelGeo: Probably follow similar behaviour as with Maps. Just add some prop tabs upgrade available or adding custom tab name as in CE is not API.

MarcelGeo (changes requested):

web-app/packages/lib/src/modules/dialog/components/ConfirmDialog.vue
xkello: Thats the fall back logic. If not variant is specified use the default logic and use logo based on severity

MarcelGeo (approved):

PR #638, opened 2026-06-12
Suppress focus halo on icon action buttons after mouse click

harminius:
This adds a one-line CSS rule on \`.icon-action-btn\` to drop the focus ring on mouse interaction while preserving it for keyboard navigation (accessibility).


MarcelGeo (changes requested):
Just try if focus is still running. after pressing tab and navigation to button.

harminius:
> Just try if focus is still running. after pressing tab and navigation to button.

image

MarcelGeo (approved):

PR #640, opened 2026-06-17
Auth fixes

varmar05:
- Session cookie gap fix: load_user now only returns active users; @auth_required checks is_active
- Bcrypt lazy rehash: configurable BCRYPT_LOG_ROUNDS, passwords rehashed on next login if rounds differ
- Account lockout: progressive tiers via LOCKOUT_POLICY env var, 423 response with AccountLockedError + added db migration: new failed_login_attempts and locked_until columns added

### DIFF
--- a/server/mergin/auth/app.py
+++ b/server/mergin/auth/app.py
@@ -61,7 +61,11 @@ def auth_required(f=None, permissions=None):
 
     @functools.wraps(f)
     def wrapped_func(*args, **kwargs):
-        if not current_user or not current_user.is_authenticated:
+        if (
+            not current_user
+            or not current_user.is_authenticated
+            or not current_user.is_active
+        ):
             return "Authentication information is missing or invalid.", 401
         if permissions:
             for check_permission in permissions:
--- a/server/mergin/auth/app.py
+++ b/server/mergin/auth/app.py
@@ -12,6 +12,7 @@ from sqlalchemy import func
 from .commands import add_commands
 from .config import Configuration
 from .models import User
+from .errors import AccountLockedError
 
 # signal for other versions to listen to
 user_account_closed = signal("user_account_closed")
@@ -98,11 +99,25 @@ def authenticate(login, password):
     else:
         query = func.lower(User.username) == func.lower(login)
     user = User.query.filter(query).one_or_none()
-    if user and user.check_password(password):
+    if user is None:
+        return None
+    needs_commit = False
+    if user.is_locked_out():
+        raise AccountLockedError(user.locked_until)
+    if user.check_password(password):
+        if user.failed_login_attempts or user.locked_until:
+            user.reset_lockout()
+            needs_commit = True
         if user.needs_rehash():
             user.assign_password(password)
+            needs_commit = True
+        if needs_commit:
             db.session.commit()
         return user
+    else:
+        user.record_failed_login()
+        db.session.commit()
+        return None
 
 
 def generate_confirmation_token(app, email, salt):
--- a/server/mergin/auth/app.py
+++ b/server/mergin/auth/app.py
@@ -91,12 +91,17 @@ def edit_profile_enabled(f):
 
 
 def authenticate(login, password):
+    from ..app import db
+
     if "@" in login:
         query = func.lower(User.email) == func.lower(login)
     else:
         query = func.lower(User.username) == func.lower(login)
     user = User.query.filter(query).one_or_none()
     if user and user.check_password(password):
+        if user.needs_rehash():
+            user.assign_password(password)
+            db.session.commit()
         return user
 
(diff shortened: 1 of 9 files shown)


server/mergin/auth/models.py
```diff
@@ -13,10 +13,20 @@
 from ..app import db
 from ..sync.models import ProjectUser
 from ..sync.utils import get_user_agent, get_ip, get_device_id, is_reserved_word
+from .errors import AccountLockedError
```
harminius: not used?

server/mergin/auth/errors.py
```diff
@@ -0,0 +1,21 @@
...
+class AccountLockedError(Exception, ResponseError):
+    code = "AccountLocked"
+    detail = "Account temporarily locked due to too many failed login attempts"
+
+    def __init__(self, locked_until: datetime.datetime):
+        self.locked_until = locked_until
+
+    def to_dict(self) -> Dict:
+        data = super().to_dict()
+        data["locked_until"] = self.locked_until.isoformat()
```
harminius: shall we make it timezone aware when it's used in client-facing public API?
we already use it when going through schemas

server/mergin/auth/models.py
```diff
@@ -64,12 +78,57 @@ def check_password(self, password):
...
 
+    def needs_rehash(self):
+        """Return True if the stored hash was generated with a different cost factor than configured."""
+        if self.passwd is None:
+            return False
+        rounds = current_app.config.get("BCRYPT_LOG_ROUNDS", 12)
+        try:
+            # bcrypt hash format: $2b$<rounds>$<salt+hash>
+            hash_rounds = int(self.passwd.split("$")[2])
+            return hash_rounds != rounds
```
harminius: `<` would be probably safer, but not a big deal

harminius (approved):
Looks good 🛡️

server/mergin/auth/controller.py
```diff
@@ -238,7 +245,10 @@ def admin_login():  # pylint: disable=W0613,W0612
     if not form.validate():
         return jsonify(form.errors), 400
 
-    user = authenticate(form.login.data, form.password.data)
+    try:
+        user = authenticate(form.login.data, form.password.data)
+    except AccountLockedError as e:
+        abort(423, f"Account temporarily locked until {e.locked_until.isoformat()}")
```
MarcelGeo: If we want to return exact time (not sure If it's compatible with security guys) - give it to attribute in AccountLockedError new attribute. I think we can show something for user in clients - received count and reset-at time.

server/mergin/auth/app.py
```diff
@@ -87,13 +92,32 @@ def wrapped_func(*args, **kwargs):
...
+        if user.failed_login_attempts or user.locked_until:
+            user.reset_lockout()
+            needs_commit = True
+        if user.needs_rehash():
+            user.assign_password(password)
+            needs_commit = True
+        if needs_commit:
+            db.session.commit()
         return user
+    else:
```
MarcelGeo, line 122: If the user attempts failing after 5 times -> wait -> is_locked_out will remove locked_until. After 60 seconds another fail will lead to next 60s ... It's like not reseting properly.

server/mergin/auth/models.py
```diff
@@ -64,12 +78,57 @@ def check_password(self, password):
...
+            return False
+
+    def is_locked_out(self) -> bool:
+        """Return True if the account is currently under a temporary lockout."""
+        if self.locked_until is None:
+            return False
+        now = datetime.datetime.utcnow()
+        if self.locked_until <= now:
+            # lockout has expired — clear it so subsequent queries see a clean state
+            self.locked_until = None
```
MarcelGeo: Is commited somewhere or it's just related to needsCommit in authenticate

MarcelGeo (changes requested):
Do we need something to do in FE/clients/mobile - login attempts logic? @tomasMizera 

@varmar05 We didn't implemented ip reputation in this mechanism -> There could be case when somebody can lock another account which he knows by email or accidentally similar emails can block each together. This could be probably handled by infra rate limiting - needs to discuss what values to adjust for current nginx config files in this repository - could be upgraded also nginx proxy. 

Flask limiter wouldn't do the same job as database locking - probably issue with redis backend again.

server/mergin/auth/controller.py
varmar05: we agreed to remove time information altogether

server/mergin/auth/models.py
varmar05: it is safe from needs_commit later perspective, but correct, I will make this method read only, not to modify the state

MarcelGeo (approved):

Issue #573, opened 2026-02-02
Support for .py python script files in Mergin projects


holesond:
As far as I know, uploading python script files (and bash scripts) to Mergin projects is not allowed since pull request #370. However, some of our projects stored on a self-hosted Mergin server (Community Edition) need to distribute python or bash scripts together with QGIS projects. The scripts help project users to automatically create new vector or raster layers on demand or process data already supplied by the users. The scripts are often tightly integrated with the projects so distributing and updating them via a different channel such as GitHub instead of Mergin would be inconvenient.

Would it be possible to add a per-project or a server-wide configuration option which would disable the "dangerous file type upload restriction"?

PR #370, opened 2025-02-06
Restrict upload for dangerous file types

harminius:
Resolves https://github.com/MerginMaps/server-private/issues/2732 

Problem:
We allow users to upload any file, even the (potentially) malicious ones :pirate_flag: 

See tickets for more details and the Pentest report

Solution:
We are more confident with using **blacklists**. :shield: 

Validate extension during project push - aka `push_start`. Abort the upload if it includes files with blacklisted extensions.
Do not rely solely on the extension check. Determine the file type from its header and block blacklisted types - once we have the file on our filesystem (in `push_finish`).

Zipping the unsupported makes it possible to upload it.

E.g. block `.exe`

Screenshot from 2025-02-06 08-08-00

Extension renaming of the unsupported file doesn't help :no_entry: 

Screenshot from 2025-02-06 08-40-48

Potential whitelists are in [the comment](https://github.com/MerginMaps/server/pull/370#issuecomment-2647315058)

~~Do not allow to sync files with extensions unless whitelisted~~

~~Whitelist approach is suggested in this PR as it is considered to be safer. However, using blacklists is also a way, [see this comment](https://github.com/MerginMaps/server/pull/370#issuecomment-2642074240) for suggested blacklists.~~

PR #643, opened 2026-06-22
Support for whitelisting extensions/mimetype

harminius:
Resolves https://github.com/MerginMaps/server/issues/573

Since #370, the server rejects "dangerous" file types (e.g. .py, .sh) on upload - but self-hosted deployments that legitimately distribute scripts alongside QGIS projects had no way to opt out.

This adds an optional, comma-separated config var to bypass file security checks:

`UPLOAD_EXTENSIONS_WHITELIST` - extra file extensions to permit, e.g. .py, .sh

image

### DIFF
--- a/deployment/community/.env.template
+++ b/deployment/community/.env.template
@@ -109,9 +109,8 @@ LOCAL_PROJECTS=/data
 
 #BLACKLIST='.mergin/, .DS_Store, .directory'  # cast=Csv()
 
-# extra file types to permit beyond the default block-list (e.g. scripts)
-#UPLOAD_EXTENSIONS_WHITELIST=''  # cast=Csv()
-#UPLOAD_MIME_TYPES_WHITELIST=''  # cast=Csv()
+# extra file extensions to permit beyond the default block-list, e.g. '.py, .sh'
+#UPLOAD_EXTENSIONS_WHITELIST=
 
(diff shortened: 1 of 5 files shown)


deployment/community/.env.template
```diff
@@ -109,6 +109,10 @@ LOCAL_PROJECTS=/data
 
 #BLACKLIST='.mergin/, .DS_Store, .directory'  # cast=Csv()
 
+# extra file types to permit beyond the default block-list (e.g. scripts)
```
MarcelGeo: I think give there just empty value without cast.

server/mergin/sync/utils.py
```diff
@@ -315,6 +315,8 @@ def is_supported_extension(filepath) -> bool:
     if check_skip_validation(filepath):
         return True
     ext = os.path.splitext(filepath)[1].lower()
+    if ext in {e.lower() for e in Configuration.UPLOAD_EXTENSIONS_WHITELIST}:
```
MarcelGeo: I think we can use check_skip_validation function for that job to unify with files or?

MarcelGeo (changes requested):

varmar05 (approved):

MarcelGeo (approved):

PR #647, opened 2026-07-01
Add audit logs

varmar05:
Introduce a pluggable audit module with NullSink default and emit() API.

Wire SQLAlchemy listeners for user and project lifecycle events, explicit emits for auth and sync endpoints (login, password, access grants/revocations, version push, soft/hard delete, restore).

Add actor_context()/request_context() helpers, device_id capture, scope_id for workspace-scoped filtering, and target_type auto-derivation.


server/mergin/auth/controller.py
```diff
@@ -47,6 +50,23 @@
...
+        return user.email, user.id
+    found = (
+        User.query.filter(
+            (func.lower(User.email) == func.lower(login_str))
+            | (func.lower(User.username) == func.lower(login_str))
+        )
+        .with_entities(User.id, User.email)
+        .first()
+    )
+    return (found.email, found.id) if found else (None, None)
```
harminius, line 67: how about `return ... else (login_str, None)` to know at least something?

server/mergin/auth/controller.py
```diff
@@ -254,10 +319,30 @@ def admin_login():  # pylint: disable=W0613,W0612
...
+                AuthEventType.USER_LOGIN_SUCCEEDED,
+                actor_id=user.id,
+                actor_email=user.email,
+                **request_context(),
+                user_id=user.id,
+                login_method="password",
+            )
             return "", 200
         else:
             abort(403, "You do not have permissions")
```
harminius, line 332: we don't want to log this?
we do log login fail..

server/mergin/auth/controller.py
```diff
@@ -472,22 +573,47 @@ def update_user(username):  # pylint: disable=W0613,W0612
...
+            target_email=user.email,
+        )
+    elif not old_active and user.active:
+        emit(
+            AuthEventType.USER_RESTORED,
+            **actor_context(),
+            user_id=user.id,
+            target_email=user.email,
+        )
+
```
harminius, line 596: we don't log superadmin assignment?

server/mergin/auth/controller.py
```diff
@@ -472,22 +573,47 @@ def update_user(username):  # pylint: disable=W0613,W0612
...
+    old_active = user.active
     form.update_obj(user)
-
-    # remove inactive since flag for ban or re-activation
     user.inactive_since = None
-
     db.session.add(user)
     db.session.commit()
+
+    if old_active and not user.active:
```
harminius, line 582: should we re-querry the `user`?

server/mergin/auth/controller.py
```diff
@@ -472,22 +573,47 @@ def update_user(username):  # pylint: disable=W0613,W0612
...
 
 @auth_required(permissions=["admin"])
 def delete_user(username):  # pylint: disable=W0613,W0612
     user = User.query.filter_by(username=username).first_or_404("User not found")
+    emit(
+        AuthEventType.USER_MARKED_FOR_DELETION,
+        **actor_context(),
+        user_id=user.id,
+        target_email=user.email,
+    )
```
harminius, line 608: shall we emit after the action?

server/mergin/sync/private_api_controller.py
```diff
@@ -99,8 +110,17 @@ def decline_project_access_request(request_id):  # noqa: E501
         project_role == ProjectRole.OWNER
         or current_user.id == access_request.requested_by
     ):
+        requester = User.query.get(access_request.requested_by)
```
harminius, line 113: can we use the new SQL syntax?

server/mergin/sync/private_api_controller.py
```diff
@@ -99,8 +110,17 @@ def decline_project_access_request(request_id):  # noqa: E501
...
         access_request.resolve(RequestStatus.DECLINED, current_user.id)
         db.session.commit()
+        emit(
+            SyncEventType.PROJECT_ACCESS_REQUEST_REJECTED,
+            **actor_context(),
+            project_id=project.id,
+            workspace_id=project.workspace_id,
+            target_email=requester.email if requester else None,
```
harminius, line 121: we should get the actor email when they are authenticated, no?
(applies also for the following emits)

server/mergin/sync/private_api_controller.py
```diff
@@ -124,7 +144,25 @@ def accept_project_access_request(request_id):
...
+            project_name=f"{project.workspace.name}/{project.name}",
+            role=permission,
+        )
+        emit(
+            SyncEventType.PROJECT_MEMBER_ADDED,
+            **actor_context(),
+            project_id=project.id,
+            workspace_id=project.workspace_id,
+            target_email=requester.email if requester else None,
+            role=permission,
```
harminius, line 164: the project name could be useful here

server/mergin/sync/private_api_controller.py
```diff
@@ -240,7 +285,16 @@ def force_project_delete(id):  # noqa: E501
...
+    emit(
+        SyncEventType.PROJECT_DELETED,
+        **actor_context(),
+        project_id=project.id,
+        workspace_id=project.workspace_id,
+        project_name=f"{project.workspace.name}/{project.name}",
+    )
+    db.session.info["audit_skip_project_update"] = True
     project.delete()
+    db.session.info.pop("audit_skip_project_update", None)
```
harminius, line 297: it would be safer to put his in `try:` and `finally:` to avoid stale info message, but that's rather a theoretical risk

server/mergin/sync/public_api_v2_controller.py
```diff
@@ -172,11 +208,21 @@ def update_project_collaborator(id, user_id):
...
+    user = User.query.get(user_id)
     project.unset_role(user_id)
     db.session.commit()
+    emit(
+        SyncEventType.PROJECT_MEMBER_DELETED,
+        **actor_context(),
+        project_id=project.id,
+        workspace_id=project.workspace_id,
+        target_email=user.email if user else None,
+        role=removed_role.value,
```
harminius, line 224: again, the project name

server/mergin/sync/tasks.py
```diff
@@ -64,8 +66,16 @@ def remove_projects_backups():
...
             break
 
+        db.session.info["audit_skip_project_update"] = True
         for p in projects:
+            emit_safe(
+                SyncEventType.PROJECT_DELETED,
+                project_id=p.id,
+                workspace_id=p.workspace_id,
+                project_name=f"{p.workspace.name}/{p.name}",
+            )
```
harminius, line 76: could we add info it was done by this celery task?

harminius (approved):
Some minor questions.

.gitignore
```diff
@@ -36,3 +36,6 @@ docker-compose.local.yml
 # SSO
 *.pem
 *.crt
+
+# Local Claude Code skills
+.claude/commands/
```
MarcelGeo, line 41: Share it with us :) If usefull.

server/mergin/tests/utils.py
```diff
@@ -4,6 +4,7 @@
 
 import json
 import shutil
+from typing import List
```
MarcelGeo, line 7: Do we need this typing If python is bumped?

server/mergin/sync/listeners.py
```diff
@@ -0,0 +1,61 @@
...
+# version push (disk_usage, latest_version, tags), operational metadata (updated,
+# storage_params, locked_until), or covered by dedicated events with their own emit
+# (removed_at/removed_by are suppressed via audit_skip_project_update instead).
+# frozenset prevents accidental mutation of module-level state.
+_SKIP = frozenset(
+    {
+        "disk_usage",
+        "latest_version",
+        "updated",
+        "storage_params",
```
MarcelGeo, line 22: I think when storage params are updated, this is like critical bug.

server/mergin/sync/listeners.py
```diff
@@ -0,0 +1,61 @@
...
+from ..audit.listeners import actor_context, field_changes, emit_safe
+from .events import SyncEventType
+from .models import Project
+
+# Fields excluded from project.updated audit events — either auto-computed on every
+# version push (disk_usage, latest_version, tags), operational metadata (updated,
+# storage_params, locked_until), or covered by dedicated events with their own emit
+# (removed_at/removed_by are suppressed via audit_skip_project_update instead).
+# frozenset prevents accidental mutation of module-level state.
+_SKIP = frozenset(
```
MarcelGeo, line 17: Do we have alternatives in other events to this columns?

rename this variable to something meaningful.

server/mergin/sync/listeners.py
```diff
@@ -0,0 +1,61 @@
...
+    {
+        "disk_usage",
+        "latest_version",
+        "updated",
+        "storage_params",
+        "tags",
+        "locked_until",
```
MarcelGeo, line 24: Yes, we are not locking projects using flask, but in future it should be critical to have it in audits

server/mergin/sync/private_api_controller.py
```diff
@@ -240,7 +285,16 @@ def force_project_delete(id):  # noqa: E501
     )
     if not project.removed_at:
         abort(400, "Failed to remove: Project is still active")
+    emit(
```
MarcelGeo, line 288: Why is this not handled in project.delete() method after commit? I Think We have multiple cases like this - project deleted, user deleted ...

server/mergin/auth/controller.py
```diff
@@ -381,6 +477,11 @@ def unlock_account(token: str):  # pylint: disable=W0613,W0612
 
     user.reset_lockout()
     db.session.commit()
+    emit(
```
MarcelGeo, line 480: Not better to add to db event? Or we do not know how to check columns in db events?

server/mergin/audit/listeners.py
```diff
@@ -0,0 +1,90 @@
...
+        if hist.has_changes():
+            old = hist.deleted[0] if hist.deleted else None
+            new = hist.added[0] if hist.added else None
+            if old != new:
+                ctx[f"old_{attr.key}"] = old
+                ctx[f"new_{attr.key}"] = new
+    return ctx
+
+
+def emit_safe(event_type, **kwargs):
```
MarcelGeo, line 79: Why not use emit_safe in controllers? We want to break endpoints if something is wrong with audit logs write?

MarcelGeo (changes requested):

server/mergin/auth/controller.py
varmar05: removed completely

server/mergin/auth/controller.py
varmar05: good idea, done

server/mergin/auth/controller.py
varmar05: added

server/mergin/auth/controller.py
varmar05: done

server/mergin/sync/private_api_controller.py
varmar05: requester here is not necessarily logged in user, can be also workspace member but in principle such user should exist if not removed meanwhile

server/mergin/sync/public_api_v2_controller.py
varmar05: added wherever we find it possible and useful

.gitignore
varmar05: it is mostly how to run tests locally with my set up / local env, nothing interesting :-)

server/mergin/tests/utils.py
varmar05: removed

server/mergin/sync/listeners.py
varmar05: this should never happen besides project deleted which has its own event (and that is a reason for being skipped) - critical bug is not a security event caused by user so I would not treat that differently

server/mergin/sync/listeners.py
varmar05: renamed

server/mergin/sync/listeners.py
varmar05: ok, removed, albeit we agreed this will be special flag only ever set by manual ops intervention

server/mergin/sync/private_api_controller.py
varmar05: good catch, moved into method

server/mergin/auth/controller.py
varmar05: db events are only used for generic object update caused by user call for update - other events have their own dedicated events rather that if-else branches in update

server/mergin/audit/listeners.py
varmar05: we definitely do not want logs to disrupt user actions so it should be always safe emit - fixed

server/mergin/auth/controller.py
varmar05: not needed, current state is held in memory

MarcelGeo (approved):

PR #656, opened 2026-07-24
Send email notification with self-service unlock link on account lockout

varmar05:
Alerts the account owner when repeated failed logins lock their account and lets them unlock it immediately instead of waiting out the lockout, via a token bound to the specific lock episode so it can't be reused or replayed across later lockouts.


web-app/packages/lib/src/modules/user/views/AccountUnlockView.vue
```diff
@@ -0,0 +1,65 @@
+<!--
+Copyright (C) Lutra Consulting Limited
+
+SPDX-License-Identifier: AGPL-3.0-only OR LicenseRef-MerginMaps-Commercial
+-->
+
+<template>
+  <app-onboarding-page>
+    <template #header>
+      <h1 class="headline-h1">Account unlock</h1>
```
MarcelGeo: Some different text.

web-app/packages/lib/src/modules/user/views/AccountUnlockView.vue
```diff
@@ -0,0 +1,65 @@
...
+  </app-onboarding-page>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+
+import AppOnboardingPage from '@/common/components/AppOnboardingPage.vue'
+import { UserApi } from '@/modules/user/userApi'
+
+export default defineComponent({
```
MarcelGeo: rewrite to <script setup> approach, as new and such simple components wold be better in <script setup lang=ts>

MarcelGeo (dismissed):

web-app/packages/lib/src/modules/user/views/AccountUnlockView.vue
varmar05: @tomasMizera any preference?

server/mergin/auth/app.py
```diff
@@ -163,3 +165,61 @@ def send_confirmation_email(app, user, url, template, header, **kwargs):
...
+    return serializer.dumps(payload, salt=app.config["SECURITY_UNLOCK_SALT"])
+
+
+def confirm_unlock_token(token, expiration=24 * 3600):
+    serializer = URLSafeTimedSerializer(current_app.config["SECRET_KEY"])
+    try:
+        payload = serializer.loads(
+            token, salt=current_app.config["SECURITY_UNLOCK_SALT"], max_age=expiration
+        )
+    except Exception:
```
harminius: this broad Exception is unavoidable?

server/mergin/tests/test_auth.py
```diff
@@ -161,6 +176,82 @@ def assert_locked():
...
 
+    # no email on successful login
+    assert send_email_mock.call_count == 2
+
+
+@patch("mergin.celery.send_email_async.apply_async")
+def test_unlock_account(send_email_mock, client, app):
+    """Test the self-service unlock-account link: valid use, reuse, natural
+    expiry, and cross-tier reuse, per the token-binding design."""
+    client.application.config["LOCKOUT_POLICY"] = "3:60,4:3600"
```
harminius: i think we prefer mocking the config variables rather than overriding them to avoid influencing other tests
not a big deal in this case though

harminius (approved):
Please add types to function arguments where missing

PR #657, opened 2026-08-04
locked account fixes

varmar05:
Use LoginHistory for account lockout. Enable all attempts to be stored in the table (needs periodic pruning to avoid storing archive data). Also DB-sync agent is now fully tracked in login_history table.

Display custom message in FE when account is locked.

### DIFF
--- a/server/migrations/community/7a095270b252_login_history_index_concurrently.py
+++ b/server/migrations/community/7a095270b252_login_history_index_concurrently.py
@@ -1,0 +1,41 @@
+"""Create login_history user/successful/timestamp index concurrently
+
+CONCURRENTLY avoids blocking writes to login_history during the build
+(every login attempt writes here), at the cost of running outside the
+migration's transaction (autocommit_block) and leaving an INVALID index
+behind on failure - upgrade()/downgrade() self-heal by dropping that first.
+
+Revision ID: 7a095270b252
+Revises: a3c8f2e1d947
+Create Date: 2026-07-24 00:00:00.000000
+
+"""
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "7a095270b252"
+down_revision = "a3c8f2e1d947"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.get_context().autocommit_block():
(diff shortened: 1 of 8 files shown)


web-app/packages/lib/src/modules/user/store.ts
```diff
@@ -238,11 +240,20 @@ export const useUserStore = defineStore('userModule', {
...
       const instanceStore = useInstanceStore()
       const formStore = useFormStore()
+      const notificationStore = useNotificationStore()
 
       try {
         await UserApi.login(payload.data)
         await instanceStore.initApp()
-      } catch (error) {
+      } catch (err) {
+        const error = err as AxiosError
```
MarcelGeo: U can use `axios.isAxiosError()` method to check if it's error from request and not from code above. Better than forcing as AxiosError

MarcelGeo (changes requested):

server/migrations/community/7a095270b252_login_history_index_concurrently.py
```diff
@@ -0,0 +1,41 @@
...
+revision = "7a095270b252"
+down_revision = "a3c8f2e1d947"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.get_context().autocommit_block():
+        op.execute(
+            "DROP INDEX CONCURRENTLY IF EXISTS ix_login_history_user_id_successful_timestamp"
```
MarcelGeo, line 27: Is it possible in alembic?

MarcelGeo (changes requested):

server/migrations/community/7a095270b252_login_history_index_concurrently.py
varmar05: It seems it is possible - I created a dedicated migration for this as it cannot be in multiquery transaction.
Worth testing though - if it turns out ok we can avoid manual db patches in future.

I'd keep this open until we verify.

varmar05: Ok, confirmed as working

MarcelGeo (approved):

PR #658, opened 2026-08-07
Apply default_version() when version is empty string

harminius:
The version variable is correctly populated for release builds, but breaks cd.dev: in the SaaS build workflow_dispatch path (non-release), VERSION is set to an empty string, and the default=get_version() fallback never fires and /config reports an empty version. That breaks client feature checks of the supported server version.

image


harminius:
Already fixed by https://github.com/MerginMaps/server-private/pull/3452/changes/14ad90c800f8615897c62e34ef0e4aa8aca72d4f

PR #663, opened 2026-08-18
Fix auth responses

varmar05:
Make sure that auth endpoints have the same timings and response codes so it is not possible to guess the error type from response itself.


MarcelGeo (approved):
@varmar05 Please adjust related tickets.

